Issue #1, opened 2020-10-13
💎 Select gems and build commented Gemfile


chrisma:
A Gemfile lists all dependencies. Ensure that the project has a good selection of the most useful gems for RoR projects.

Last course's Gemfile is a good starting point: https://github.com/hpi-swt2/lecture-portal/blob/dev/Gemfile

- [x] Make sure all gems are up-to-date
- [x] Make sure all gems are commented, including their benefit and links to websites
- [x] Ensure gems work with the selected Rails and Ruby version and throw no runtime errors in production mode

LucPrestin:
I uploaded a Gemfile to the branch skeleton/#1-gem-selection

If you miss one, let me know

chrisma:
Thanks! I’ll have a look! Anything so should pay special attention to?

LucPrestin:
I don't think so. I just want to be sure nothing is missing.

LucPrestin:
In fact: there is something I wanted to ask.

I remember from last year, that you insisted on ruby 2.5.1. 
The ruby version, pacman installs by default is ruby 2.7.1, so I set that as the ruby version. Was there a particular reason to insist on ruby 2.5.1 last year?

chrisma:
> Was there a particular reason to insist on ruby 2.5.1 last year?

Was that possibly the latest stable version last year? The current latest stable version is 2.7.2. I have added that version to the Gemfile. We should also add it to a `.ruby_version` file. **EDIT:** Done. 

Unfortunately, distribution package managers are not always up-to-date. I recommend using [rbenv](https://github.com/rbenv/rbenv-installer#rbenv-installer) as it also allows switching between different ruby versions.

chrisma:
Here is a **default** Gemfile created with Rails 6 using `rails new blog --skip-spring --skip-listen`

```ruby
source 'https://rubygems.org'
git_source(:github) { |repo| "https://github.com/#{repo}.git" }

ruby '2.7.2'

# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
gem 'rails', '~> 6.0.3', '>= 6.0.3.4'
# Use sqlite3 as the database for Active Record
gem 'sqlite3', '~> 1.4'
# Use Puma as the app server
gem 'puma', '~> 4.1'
# Use SCSS for stylesheets
gem 'sass-rails', '>= 6'
# Transpile app-like JavaScript. Read more: https://github.com/rails/webpacker
[...]

PR #12, opened 2020-10-19
Add Gemfile

LucPrestin:
Resolves Issue #1 

- Add a Gemfile with the basic needs for the application


Gemfile
```diff
@@ -0,0 +1,100 @@
+source "https://rubygems.org"
+
+git_source(:github) {|repo_name| "https://github.com/#{repo_name}.git" }
+
+# ruby version
+ruby '2.7.1'
```
chrisma: 2.7.2 is latest, no? https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/

chrisma (approved):
How different is this Gemfile from the standard one that is generated using `rails new`? Any significant diffs we should be wary of? *EDIT: Updated Gemfile with Rails 6 defaults.*

Gemfile
```diff
@@ -0,0 +1,100 @@
...
+  gem 'rubocop-rspec', '~> 1.43', '>= 1.43.2', require: false
+  # Extension for Rubocop https://github.com/rubocop-hq/rubocop-performance
+  gem 'rubocop-performance', '~> 1.8', '>= 1.8.1', require: false
+end
+
+group :development do
+  # Access an interactive console on exception pages or by calling 'console' anywhere in the code.
+  gem 'web-console', '~> 4.0', '>= 4.0.4' # https://github.com/rails/web-console
+  # The Listen gem listens to file modifications and notifies you about the changes.
+  gem 'listen', '~> 3.2', '>= 3.2.1' # https://github.com/guard/listen
```
chrisma: Students might be using the WSL: The RoR tutorial notes:
> If you're using Windows Subsystem for Linux then there are currently some limitations on file system notifications that mean you should disable the spring and listen gems which you can do by running rails new blog --skip-spring --skip-listen.

Are spring and listen required (for other gems to work)?

chrisma (approved):

Gemfile
```diff
@@ -0,0 +1,100 @@
...
+  gem 'webdrivers', '~> 4.4', '>= 4.4.1'
+  # Port of Perl's Data::Faker library that generates fake data
+  gem 'faker', '~> 2.14'
+  # Provides one-liners to test common rails functionality, that, if written by hand, would be much longer
+  gem 'shoulda-matchers', '~> 4.4', '>= 4.4.1'
+end
+
+group :production do
+  # production database runs on postgres
+  gem 'pg', '~> 1.2', '>= 1.2.3'
```
chrisma: Should we run production on SQLite as well this time to prevent incompatibilities between dev & prod? We most likely don't need the power of Postgres (especially at the beginning). Also, one fewer install. I would vote for trying it.

chrisma (approved):

Gemfile
```diff
@@ -0,0 +1,100 @@
...
+# The application framework. https://github.com/rails/rails
+gem 'rails', '~> 6.0', '>= 6.0.3.4'
+# Use sqlite3 as the database for Active Record. https://www.sqlite.org/index.html
+gem 'sqlite3', '~> 1.4', '>= 1.4.2'
+# Use Puma as the app server. https://github.com/puma/puma
+gem 'puma', '~> 5.0', '>= 5.0.2'
+# Use Uglifier as compressor for JavaScript assets. https://github.com/lautis/uglifier
+gem 'uglifier', '~> 4.2'
+# See https://github.com/rails/execjs#readme for more supported runtimes
+gem 'mini_racer', '~> 0.3.1'
```
chrisma: Is this still required in Rails 6? Is it included in the default Gemfile that is created for a new Rails 6 project? If I understood correctly, Node.js is required now anyways, so `execjs` could use that instead of `mini_racer`

chrisma (approved):

Gemfile
chrisma: Updated

chrisma:
Here is a **default** Gemfile created with Rails 6 and without `spring` and `listen`: 

```ruby
source 'https://rubygems.org'
git_source(:github) { |repo| "https://github.com/#{repo}.git" }

ruby '2.7.2'

# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
gem 'rails', '~> 6.0.3', '>= 6.0.3.4'
# Use sqlite3 as the database for Active Record
gem 'sqlite3', '~> 1.4'
# Use Puma as the app server
gem 'puma', '~> 4.1'
# Use SCSS for stylesheets
gem 'sass-rails', '>= 6'
# Transpile app-like JavaScript. Read more: https://github.com/rails/webpacker
gem 'webpacker', '~> 4.0'
# Turbolinks makes navigating your web application faster. Read more: https://github.com/turbolinks/turbolinks
gem 'turbolinks', '~> 5'
# Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
gem 'jbuilder', '~> 2.7'
# Use Redis adapter to run Action Cable in production
# gem 'redis', '~> 4.0'
# Use Active Model has_secure_password
# gem 'bcrypt', '~> 3.1.7'

# Use Active Storage variant
# gem 'image_processing', '~> 1.2'

# Reduces boot times through caching; required in config/boot.rb
gem 'bootsnap', '>= 1.4.2', require: false

group :development, :test do
  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
  gem 'byebug', platforms: [:mri, :mingw, :x64_mingw]
end

group :development do
  # Access an interactive console on exception pages or by calling 'console' anywhere in the code.
  gem 'web-console', '>= 3.3.0'
end

group :test do
  # Adds support for Capybara system testing and selenium driver
  gem 'capybara', '>= 2.15'
  gem 'selenium-webdriver'
  # Easy installation and use of web drivers to run system tests with browsers
  gem 'webdrivers'
end

# Windows does not include zoneinfo files, so bundle the tzinfo-data gem
gem 'tzinfo-data', platforms: [:mingw, :mswin, :x64_mingw, :jruby]
```

I don't know what `webpacker` actually does, but I've added it, as it might be required by other Rails components.
**EDIT:** `webpacker` is new in Rails 6: https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#using-webpacker

Gemfile
chrisma: I think the answer is: no. I've removed them

Gemfile
chrisma: And it's gone!

Gemfile
chrisma: The default Rails 6 Gemfile does not contain `mini_racer`, so I've removed it.

Issue #2, opened 2020-10-13
🏗️ Setup automated Heroku deploy


chrisma:
As teams will be remote much of the time, a deployed version of the application that is automatically updated seems even more important than previously and should be available from the start of the course.

I suggest using [Heroku](https://www.heroku.com/) again, as it has worked well for RoR apps in the past (but I'm also open to alternatives). There's no reason not to have deployments on multiple providers.

- [x] Automated deploy of **dev** branch, after successful tests, to Heroku (using [GitHub Actions](https://github.com/marketplace/actions/deploy-to-heroku))
- [x] Automated deploy of **main** branch, after successful tests, to Heroku (using [GitHub Actions](https://github.com/marketplace/actions/deploy-to-heroku))
- [x] Add Heroku badges to README
- [x] Add links to deployments to README
- [x] Add information on how to find detailed information on last test / deployment run in README

LucPrestin:
- [x] Add the app on Heroku
- [x] Add Heroku token to the github secrets
- [x] Add app name and secret token to the workflow file

LucPrestin:
Does the chair have a Heroku-Account that we can use for the deployment or do I have to create one?

chrisma:
> Does the chair have a Heroku-Account that we can use for the deployment or do I have to create one?

You have mail :)

LucPrestin:
What are Heroku-Badges? I could not find any information on that.

chrisma:
> What are Heroku-Badges? I could not find any information on that.

Apologies, that is bad wording on my part... I meant these (in the README):
![image](https://user-images.githubusercontent.com/1652117/96734938-52915380-13bb-11eb-9a68-882ca511bf0a.png)

LucPrestin:
> > What are Heroku-Badges? I could not find any information on that.
> 
> Apologies, that is bad wording on my part... I meant these (in the README):
> ![image](https://user-images.githubusercontent.com/1652117/96734938-52915380-13bb-11eb-9a68-882ca511bf0a.png)

I see. I will add them.

LucPrestin:
I added both badges, but the build badge will only work when the workflow file gets to the master branch, as GitHub only generates the workflow when it is.

chrisma:
> I added both badges, but the build badge will only work when the workflow file gets to the master branch

Let's do it (and merge the preparation branch into main to test this out)!

LucPrestin:
[...]

Issue #13, opened 2020-10-20
⚙️ Document dependencies in README


chrisma:
The following things are required: https://guides.rubyonrails.org/getting_started.html#creating-a-new-rails-project-installing-rails

![image](https://user-images.githubusercontent.com/1652117/96577471-f56ea280-12d3-11eb-9cc2-6d9f54832e4a.png)

It's true:

Node ![image](https://user-images.githubusercontent.com/1652117/96577360-ce17d580-12d3-11eb-8b1d-2a204b233249.png)

chrisma:
Hmmm... If I'm understanding it correctly, Node.js and Yarn are only required for webpacker, which simply bundles assets (https://webpack.js.org/). This job can also be done via the previous Rails Assets Pipeline, maybe not as nicely?
However, it seems neater to just stick with the Asset Pipeline (with internal, gem dependencies incl. `mini_racer`) than having everyone install node and yarn manually.

chrisma:
Actually, this whole Webpack(er) things seems fairly dumb, imho

> Like Webpack, Sprockets is an asset pipeline, which means it takes assets files as input (Javascript, CSS, images, etc.) and processes them to produce an output in the desired format.
As of Rails 6, Webpack(er) replaces Sprockets as the new standard for writing Javascript in your Rails applications. However, Sprockets is still the default way of adding CSS to your applications.
-- https://blog.capsens.eu/how-to-write-javascript-in-rails-6-webpacker-yarn-and-sprockets-cdf990387463

We should try to stick to the Rails (5) Asset Pipeline using Sprockets for the project, which handles all assets. We don't care if ES6 is maybe not handled well, just don't write that. We're still including jQuery over here... I don't think we necessarily need the power (and complexity) of webpack

chrisma:
Maybe StackOverflow knows more: https://stackoverflow.com/questions/64466331/is-it-possible-advisable-to-run-rails-6-without-webpacker

chrisma:
I've also opened an issue on the rails project: https://github.com/rails/rails/issues/40438

chrisma:
The intro exercise will definitely not use webpacker and will continue to rely on Sprockets and the Rails Asset Pipeline. It doesn't need JS and thus doesn't really care about node.js and yarn dependencies.

It's definitely possible to rely on the Asset Pipeline in Rails 6 projects, I've tried it out here: https://github.com/chrisma/rails6-github-actions/commits/main

[...]

PR #15, opened 2020-10-21
Skeleton/#2 automated heroku deploy

LucPrestin:
Resolves issue #2 

- Add Github Action for CI and CD
- Update README


.github/workflows/CI_CD.yml
```diff
@@ -0,0 +1,84 @@
+name: CI/CD
+
+on:
+  push:
+    branches: [ main, dev ]
+  pull_request:
```
chrisma, line 6: We should probably not deploy PRs to Heroku by default, as we have limited Heroku Dyno hours available (Even deploying *main* and *dev* on the same account might lead to issues later, we should keep an eye on that)

.github/workflows/CI_CD.yml
```diff
@@ -0,0 +1,84 @@
...
+    branches: [ main, dev ]
+
+jobs:
+  CI:
+    runs-on: ubuntu-latest
+
+    services:
+      db:
+        image: postgres:11
```
chrisma: Are we using pg? I would argue that we'll stick to SQLite3 for both *dev* and *prod*, for the sake of simplicity. I don't foresee this app needing the power of Postgres anytime soon.

I removed pg from the Gemfile... If pg is to be used, it would need to be added there as well: https://github.com/hpi-swt2/rolodex-portal/blob/preparation/Gemfile

.github/workflows/CI_CD.yml
```diff
@@ -0,0 +1,84 @@
...
+          --health-retries 5
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      
+      - name: Install Ruby
+        uses: ruby/setup-ruby@v1
+        with: 
+          ruby-version: 2.7.1
```
chrisma: 2.7.2, please

.github/workflows/CI_CD.yml
```diff
@@ -0,0 +1,84 @@
...
+
+      - name: Install Postgres
```
chrisma: see above

.github/workflows/CI_CD.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: sudo apt-get -y install libpq-dev
+
+      - name: Cache Ruby Gems
```
chrisma, line 25: Awesome!

.github/workflows/CI_CD.yml
```diff
@@ -0,0 +1,84 @@
...
+          path: vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: ${{ runner.os }}-gems-
+      
+      - name: Install Ruby Gems
+        run: |
+          bundle config path vendor/bundle
+          bundle install --jobs 4 --retry 3
+          
+      - name: Install yarn packages
```
chrisma, line 37: I hope to get rid of webpack(er) and yarn, then this can go as well. Ref #13

chrisma (changes requested):

.github/workflows/CI_CD.yml
LucPrestin: But we want to apply CI to PRs right? If so, I will add a safeguard below, in the deployment job that catches PRs and only deploys on pushes. 

What do you mean by saying "We should keep an eye on that"? Just deploy both main and dev and see what happens or deploy only main for safety?

.github/workflows/CI_CD.yml
LucPrestin: Sure, missed that one.

.github/workflows/CI_CD.yml
LucPrestin: I did not realize, you removed Postgres. Then, of course, it does not make sense to use that for the CI-pipeline.

.github/workflows/CI_CD.yml
chrisma: > But we want to apply CI to PRs right?

Yes, please :)

> If so, I will add a safeguard below, in the deployment job that catches PRs and only deploys on pushes.

That'd be great

> What do you mean by saying "We should keep an eye on that"?

Check on Heroku to see what percentage of free hours are used up, e.g. in the middle of the month.

> Just deploy both main and dev and see what happens or deploy only main for safety?

Let's deploy both. We'll see what happens

README.md
```diff
@@ -1,2 +1,11 @@
-# rolodex-portal
+# rolodex-portal ![build](https://github.com/hpi-swt2/rolodex-portal/workflows/CI_CD/badge.svg?branch=main) [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](https://opensource.org/licenses/MIT)
```
chrisma: Would an additional `?branch=dev` also work? That'd be neat

.github/workflows/CI_CD.yml
LucPrestin: Services in GitHub Actions are run in Docker containers. The "image" key defines, what kind of image is supposed to be loaded to the docker container. That can either be an image name from the Docker Hub or a registry name (whatever is meant by the latter) [https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idservices). 

Postgres published an official docker image that can be used as how I did it in the workflow file. Unfortunately, I could not find anyone using SQLite3 in a GitHub Action and there is no official SQLite3 image on Docker Hub, just some community images. 

Because of that, I'm not sure how to apply SQLite3 to a GitHub Action (this is only the second time, I try to configure a GitHub Action).

So what do we want to do? Use pg because it is easier to put in the workflow, use a community version of SQLite3, write an own image file, or do something I haven't thought of yet?

chrisma: Thanks for investigating. Quite a nice feature of GH Actions, that external services (such as pg in our case) run in a separate Docker container!
[The documentation](https://docs.github.com/en/free-pro-team@latest/actions/guides/about-service-containers) says:
> Service containers are Docker containers that provide a simple and portable way for you to host services that you might need to test or operate your application in a workflow.

My argument would be that due to [the nature of SQLite](https://en.wikipedia.org/wiki/SQLite), no separate service/Docker image is necessary and that SQLite can be installed like any other gem dependency. This would mean getting rid of the `services` part of the yaml and letting the `bundle install` command handle everything (possibly prededed by a `sudo apt install libsqlite3-dev`, if that's not already available). Would you give that a shot?

LucPrestin: done

chrisma (approved):

Issue #4, opened 2020-10-13
🦴 Generate application skeleton


chrisma:
> ⚠ Before generating code using RoR generators, all gems that should be part of the project need to be included in the Gemfile. This ensures that all changes that gems make to the generators (e.g. a Bootstrap gem automatically generates styled views) are included.

- [x] Generate base models (contact, note, optional category) and connections between them
- [x] Document all (generate) commands that were issued to get to the initial state, so that when we significantly change the Gemfile, we can simply recreate the entire app, without having to make manual fixes. Create a shell script?
- [x] Ensure the app deploys to Heroku, see #2 
- [x] Ensure that no warnings or errors are thrown during runtime/startup

chrisma:
Now also deployed to Heroku

PR #16, opened 2020-10-23
Skeleton/#4 generate application skeleton

LucPrestin:
Resolves issue #4 

- generate skeleton code with the classes "user" and "note". They currently have a 1:n relationship.
- the steps, that generated the code are documented in the file preparation_steps.sh

TODO:
- [x] Add model class diagram to README
- [x] Add how to generate a class diagram of the models to the README
- [x] Fix root route (currently an error is thrown when localhost:3000 is opened)
- [x] The tests currently don't run successfully -> find out why and fix it

Questions:
- Currently, the class diagram (as generated by RubyMine) looks like this: 

class_diagram. 

Which has an aggregation rather than an association between the two model classes. This lead me to look at the schema.rb file. There, I found the line 
image which confused me even more. Shouldn't the note entries have a foreign key relating to a user and not the other way around? So did I mix something up when generating the classes or does RubyMine show me nonsense?


.browserslistrc
```diff
@@ -0,0 +1 @@
+defaults
```
chrisma, line 1: Do we need this?

Rakefile
```diff
@@ -0,0 +1,6 @@
+# Add your own tasks in files placed in lib/tasks ending in .rake,
+# for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
+
+require_relative 'config/application'
+
+Rails.application.load_tasks
```
chrisma, line 6: Hmmm... in its current form this Rakefile doesn't do much, as it only loads tasks from bin/rails. Do we need Rake tasks? If not, might be nice to keep the root directory somewhat clean

app/assets/stylesheets/notes.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the notes controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: https://sass-lang.com/
```
chrisma, line 3: We should add some SCSS styles here (try to use some SCSS feature), to test out whether these get compiled into a final CSS file in production. Feel free to make the design terrible (pink, green, whatever), so that it HAS to be changed.

app/assets/stylesheets/scaffolds.scss
```diff
@@ -0,0 +1,65 @@
...
+th {
+  padding-bottom: 5px; }
+
+td {
+  padding: 0 5px 7px; }
+
+div.field,
+div.actions {
+  margin-bottom: 10px; }
+
```
chrisma: Everything above line 35 seems odd to me, presumably these are Rails defaults. But we want Bootstrap defaults! I assume Bootstrap also defines sensible defaults for these elements.

app/controllers/notes_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  # GET /notes/1/edit
+  def edit
+  end
+
+  # POST /notes
+  def create
+    @note = Note.new(note_params)
+
+    if @note.save
+      redirect_to @note, notice: 'Note was successfully created.'
```
chrisma: should we put these (as parameterized) strings into an I18n file? They will get used a lot.... And it might be a good intro to I18n concepts.

app/models/user.rb
```diff
@@ -0,0 +1,7 @@
+class User < ApplicationRecord
+  # Include default devise modules. Others available are:
+  # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
+  devise :database_authenticatable, :registerable,
+         :recoverable, :rememberable, :validatable
+  has_many :notes, dependent: :delete_all
```
chrisma, line 9: Please add a note + link to what `dependent: :delete_all` does

app/models/user.rb
```diff
@@ -0,0 +1,7 @@
+class User < ApplicationRecord
+  # Include default devise modules. Others available are:
+  # :confirmable, :lockable, :timeoutable, :trackable and :omniauthable
+  devise :database_authenticatable, :registerable,
```
chrisma, line 5: Link to devise docs?

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,21 @@
...
+
+    <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
+    <%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload' %>
+  </head>
+
+  <body>
+    <div class="container-fluid">
+      <%= yield %>
+    </div>
+    <div class="container-fluid">
```
chrisma: Shouldn't notices and alerts be *above* the main container?

app/views/notes/_form.html.erb
```diff
@@ -0,0 +1,32 @@
+<%= form_with(model: note, local: true) do |form| %>
+  <% if note.errors.any? %>
+    <div id="error_explanation">
+      <h2><%= pluralize(note.errors.count, "error") %> prohibited this note from being saved:</h2>
```
chrisma: `<%= pluralize(note.errors.count, "error") %> prohibited this note from being saved:` could also become a parameterized I18n string

app/views/notes/index.html.erb
```diff
@@ -0,0 +1,31 @@
...
+
+  <tbody>
+    <% @notes.each do |note| %>
+      <tr>
+        <td><%= note.title %></td>
+        <td><%= note.content %></td>
+        <td><%= note.user_id %></td>
+        <td><%= link_to 'Show', note %></td>
+        <td><%= link_to 'Edit', edit_note_path(note) %></td>
+        <td><%= link_to 'Destroy', note, method: :delete, data: { confirm: 'Are you sure?' } %></td>
```
chrisma: `Are you sure?` -> I18n

app/views/notes/index.html.erb
```diff
@@ -0,0 +1,31 @@
+<p id="notice"><%= notice %></p>
```
chrisma: `<%= notice %>` is already shown in `app/views/layouts/application.html.erb`, right? Wouldn't that show twice then?

babel.config.js
```diff
@@ -0,0 +1,72 @@
+module.exports = function(api) {
```
chrisma, line 1: What a mess... wow. Any chance of adding some explanatory comments?

config/locales/en.yml
```diff
@@ -0,0 +1,33 @@
...
+# Instead, surround them with single quotes.
+#
+# en:
+#   'true': 'foo'
+#
+# To learn more, please read the Rails Internationalization guide
+# available at https://guides.rubyonrails.org/i18n.html.
+
+en:
+  hello: "Hello world"
```
chrisma: Don't think we'll need this anymore :)

config/routes.rb
```diff
@@ -0,0 +1,6 @@
+Rails.application.routes.draw do
+  resources :notes
+  devise_for :users
```
chrisma, line 5: link to devise docs

log/development.log
```diff
@@ -0,0 +1,26 @@
+  [1m[35m (1.7ms)[0m  [1m[34mSELECT sqlite_version(*)[0m
```
chrisma: Don't see a reason to commit this. add to `.gitignore`?

spec/factories/notes.rb
```diff
@@ -0,0 +1,7 @@
+FactoryBot.define do
```
chrisma, line 1: Want to use the faker gem?

spec/helpers/notes_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'rails_helper'
```
chrisma: We have no notes helpers, right? So we probably don't need any tests.

spec/models/note_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Note, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
chrisma: Let's add some super basic test here

spec/models/user_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe User, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
chrisma: super basic test using a factory, please, so that there is a good baseline.

spec/requests/notes_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+#
+# It only uses APIs available in rails and/or rspec-rails. There are a number
+# of tools you can use to make these specs even more expressive, but we're
+# sticking to rails and rspec-rails APIs to keep things simple and stable.
+
+RSpec.describe "/notes", type: :request do
+  # Note. As you add validations to Note, be sure to
+  # adjust the attributes here as well.
+  let(:valid_attributes) {
+    skip("Add a hash of attributes valid for your model")
```
chrisma: Check out `attributes_for` of factory_bot. Useful here! https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#build-strategies

spec/views/notes/index.html.erb_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe "notes/index", type: :view do
+  before(:each) do
+    assign(:notes, [
+      Note.create!(
```
chrisma: factory bot please

spec/views/notes/new.html.erb_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'rails_helper'
+
+RSpec.describe "notes/new", type: :view do
+  before(:each) do
+    assign(:note, Note.new(
```
chrisma: factory bot, please

spec/views/notes/show.html.erb_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe "notes/show", type: :view do
```
chrisma, line 3: I consider view tests that check for specific strings on a page kind of useless, as they get stale extremely quickly. I'd vote to simply remove them. What do you think?

test/application_system_test_case.rb
```diff
@@ -0,0 +1,5 @@
+require "test_helper"
+
+class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
+  driven_by :selenium, using: :chrome, screen_size: [1400, 1400]
```
chrisma, line 4: We should write a sample integration test using Capybara methods!

chrisma (changes requested):

.browserslistrc
LucPrestin: I don't know. This is a file that rails auto-generated. I assumed that Rails did this for a reason.

Rakefile
LucPrestin: See above

.browserslistrc
chrisma: Hmm... would you do some research to find the reason? Would be interesting :)

Rakefile
chrisma: I'd remove this for now, I don't think we need special rake tasks right now

.browserslistrc
LucPrestin: > With @babel/preset-env you can define which browsers your app supports. You do that by editing the .browserlistrc. Most tutorials and docs recommend to start with a generic configuration that supports most commonly used browsers:

[https://blog.jakoblind.no/babel-preset-env-custom-browserlistrc/](https://blog.jakoblind.no/babel-preset-env-custom-browserlistrc/)

This seems to be a babel-thing.

app/assets/stylesheets/scaffolds.scss
LucPrestin: I looked at last year's project and there were the same values. It could be that we used rails defaults back then too. But it did not seem to bother anyone then. So maybe it's just fine?

LucPrestin: The problem with the application not looking like bootstrap was that the application stylesheet did not import bootstrap. I fixed that, so these defaults should be fine, as I mentioned above

app/views/notes/index.html.erb
LucPrestin: Yep. I removed it in the app/view/notes/show.html.erb too for the same reason

spec/factories/notes.rb
LucPrestin: I don't know if that's really needed here. If I understood it correctly, the test we write for the students should be fairly simple just so that they can see how tests are written. Therefore I don't see the necessity to include the complexity of a faker-generated resource.

spec/views/notes/show.html.erb_spec.rb
LucPrestin: I'd suggest instead of checking for specific strings, we check for the values of the assigned note.

Rakefile
LucPrestin: db:migrate aborts when no rakefile exists. So I had to add it back into the code.

app/models/user.rb
chrisma: Nice, thanks!

log/development.log
```diff
@@ -0,0 +1,26 @@
+  [1m[35m (1.0ms)[0m  [1m[34mSELECT sqlite_version(*)[0m
```
chrisma, line 1: Probably don't need to commit this file as well as `db/development.sqlite3`. They can go into the `.gitignore`

spec/factories/notes.rb
chrisma: Fair, thanks for the answer.

spec/views/notes/show.html.erb_spec.rb
chrisma: Fixed

log/development.log
LucPrestin: Actually, it already is in the ```.gitignore```. For some reason, my git instance seems to ignore that ignore rule.

babel.config.js
LucPrestin: This was, again, autogenerated by rails. I don't know why babel is configured like it is and thus can only guess what it does.

But I don't recall having touched this in last year's project, so maybe it's just fine if we say something like "This is the configuration for babel \<link to wiki what babel is\>. For further information look at \<config documentation\>"?

app/assets/stylesheets/notes.scss
LucPrestin: I added some styling with basic scss elements. Is that enough or did you have some more complicated features in mind that should be used?

test/application_system_test_case.rb
LucPrestin: I used the rails command ```rails generate integration_test users``` and it generated the file ```spec/requests/users_spec.rb``` so I guess this is where rspec wants to have its integration tests. I added one example test for the sign-up process there.

chrisma:
Thanks for the additional commits @LucPrestin ! I'll have a look soon :)

test/application_system_test_case.rb
chrisma: Hmm... I would not expect feature/integration tests in the `requests` folder.

chrisma: I think I agree with [this](https://www.codewithjason.com/rails-integration-tests-rspec-capybara/#whats-the-difference)

image

chrisma: So either `feature` (the RSpec term) or `integration` or  `endtoend` would work for me

chrisma: Request specs are a specific type of RSpec test: https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec

PR #46, opened 2020-12-01
Implemented user table on homepage #22

### LINKED ISSUE #22: View list of all accounts
As a User, I'd like to see a list of all accounts at the home screen, in order to have an overview of all people using the platform.

## ✅ Conditions of satisfaction

- [x] Should be visible in the Home-Screen
- [x] Should show a table, with one row per account
- [x] In a row should be shown the email address of the account
- [x] Should show all existing accounts

SanJSp:
Fixes #22 

We have implemented that all users are displayed on the homepage. Additionally the users are displayed on an index page. Tests for the user/index are included, as testing on the homepage was not possible. 

Looking forward to your review.


Jostafarr:
user table and index page are there but the show action does not work and with that, it is impossible to visit the user edit page.

PatriciaSowa:
Tests of homepage are know included. Removed user indexpage.

PatriciaSowa:
grafik
So sieht die neue Homepage nun aus.

Jostafarr (approved):
everything works as expected.

inadicis (approved):
Nice Job. Thank's for the image file! (next time try to show with some more data)

PR #51, opened 2020-12-03
Update bundle installation instructions: Replace deprecated `--without` flag

LinqLover:
Do no longer depend on the deprecated `--without` flag, which gives the following warning at the moment:

> [DEPRECATED] The `--without` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set without 'production'`, and stop using this flag


chrisma:
Thanks!

Further info:
* https://github.com/rubygems/bundler/blob/d4993be66fa2e76b3ca00ea56a51ecab5478b726/UPGRADING.md#cli-deprecations
* capistrano/bundler#115

PR #54, opened 2020-12-04
Implement private note on contacts

### LINKED ISSUE #33: Direct option to create private note on contacts
As a user, I can only see notes that I have created for other users.
- [ ] In the notes list, I can only see notes that I have created for other users
- [ ] No one else can see these notes

Deferred because of unmet dependencies:
- add button in contact list (depends on #26)
- display note on profile of that user (depends on #30)

LinqLover:
Closes #33.

Roadmap:
- [x] Write acceptance test
- [x] Update database
- [x] Update controller
- [x] Write unit tests
- [x] Write/fix controller tests
- Update documentation in `README.md` (already done)
- [x] Code Review
  * Please see the unresolved conversations below!
  * Once everything is set to your satisfaction, **please squash when merging.**
  * Open TODO comments in source


rheaSPK:
Es besteht sowohl eine Verbindung zwischen dem Nutzer, der die Note geschrieben hat, als auch den Nutzern, die der Note assigned wurde. Wozu ist dann noch eine Abwandlung der Datenbankstruktur notwendig?

LinqLover:
GitHub is English, please :-)

> Es besteht sowohl eine Verbindung zwischen dem Nutzer, der die Note geschrieben hat

This connection is not modeled in the database until today. All we have in the present schema is this:

https://github.com/hpi-swt2/connections-portal/blob/725ba715be5edb8c92c85dd6a001b214b4ae0c93/db/schema.rb#L15-L22

Conceptionally, we will need to add another user ID for the "owning" user of the note (in contrary to the already existing ID of the user which the note is about). Btw, having a better metaphor for these rules could be helpful, do you have any ideas how we could name these two IDs? :-)

LinqLover:
Yahoo, first merge conflict! 🎉

db/schema.rb
```diff
@@ -10,14 +10,16 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2020_10_30_111940) do
+ActiveRecord::Schema.define(version: 2020_12_11_160500) do
```
LinqLover: Please review, this is my first conflict resolutions for `schema.rb`.

LinqLover:
@rheaSPK Aah! The missing tests are failing because `current_user` is not set within the model tests' execution. VSC Test Explorer did not show the exact line where the error occurred, but the CI has shed like on the dark. **Next steps:** Find out how to mock `current_user`.

.vagrant/machines/default/virtualbox/private_key
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
LinqLover: Woah! @rheaSPK I'd really suggest you to take care of your credentials and reset them now ...

LinqLover: Also, please update your global `.gitignore` according to https://github.com/hpi-swt2/connections-portal/pull/56#issuecomment-738951815.

spec/rails_helper.rb
```diff
@@ -68,4 +68,5 @@
   config.include Devise::Test::ControllerHelpers, type: :view
   # https://github.com/heartcombo/devise#integration-tests
   config.include Devise::Test::IntegrationHelpers, type: :feature
+  config.include Devise::Test::IntegrationHelpers, type: :request
```
LinqLover: What does this do?

spec/features/notes_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+describe "Notes", type: :feature do
+  before(:each) do
+    @alice = User.new(id: 1, email: "alice@example.com", password: "1234567")
+    assert @alice.save
+    @bob = User.new(id: 2, email: "bob@example.com", password: "9876543")
+    assert @bob.save
+    login(@alice)
+  end
+
+  # TODO: Split up
```
LinqLover, line 12: TODO - how can we split this up?

db/migrate/20201211160500_add_creator_user_to_notes.rb
```diff
@@ -0,0 +1,6 @@
+class AddCreatorUserToNotes < ActiveRecord::Migration[6.0]
+  def change
+    Note.connection.execute('DELETE FROM notes')
+    add_reference :notes, :creator_user, null: false, foreign_key: { to_table: :users }
```
LinqLover: My local linter adds the following warning here: `convention:Rails/NotNullColumn` Is this a warning we should care about?

app/controllers/notes_controller.rb
```diff
@@ -20,6 +20,7 @@ def edit; end
   # POST /notes
   def create
     @note = Note.new(note_params)
+    @note.creator_user_id = current_user.id
```
Alexander-Dubrawski, line 23: When I see it correctly we have a `has many` association between `User` and `Notes` (https://guides.rubyonrails.org/association_basics.html#the-has-many-association). Normally we would create a new `Note` for a `User` with `@note = User.notes.create(note_params)` (https://guides.rubyonrails.org/association_basics.html#why-associations-questionmark). Why are we not doing it here? 

We could achieve it with nested Ressources (https://guides.rubyonrails.org/routing.html#nested-resources).

Alexander-Dubrawski (commented):
Thank's for your work. In the future, I would find it quite helpful to have a PR description with more information in it. For example what exactly you implemented (Added tests, implemented the association, ...) and if helpful how you did it 🙃 .

I just found one point where I had a question.

app/controllers/notes_controller.rb
LinqLover: Thanks for the pointer to nested resources! I haven't met this concept so far so I do not know whether you should use it as often as possible or whether it is just another optional possibility. From a logical perspective, it might make sense to nest the notes after the users, but on the other hand, every user should only be able to see his own notes, so the extra route `/users/<user_id_that_must_be_the_current_user_id_and_nothing_else>/notes` feels kind of redundant. Do you see my point? Do you know more about the conventions for this concept?

app/controllers/notes_controller.rb
```diff
@@ -3,7 +3,7 @@ class NotesController < ApplicationController
 
   # GET /notes
   def index
-    @notes = Note.all
+    @notes = Note.where('creator_user_id = ?', current_user.id)
```
chrisma: How about `current_user.notes`?

app/controllers/notes_controller.rb
chrisma: `@note.creator = current_user` should be supported, no?

Or you might also consider adding the current_user of a request to `note_params` if this is something that only logged in users (where `current_user` is set by [devise ](https://github.com/heartcombo/devise#controller-filters-and-helpers)) can do.

db/migrate/20201211160500_add_creator_user_to_notes.rb
chrisma: Here are some examples from the [documentation](https://rubocop.readthedocs.io/projects/rails/en/stable/cops_rails/#railsnotnullcolumn), providing a rationale for this rule:

```Ruby
# bad
add_column :users, :name, :string, null: false
add_reference :products, :category, null: false

# good
add_column :users, :name, :string, null: true
add_column :users, :name, :string, null: false, default: ''
add_reference :products, :category
add_reference :products, :category, null: false, default: 1
```

db/schema.rb
chrisma: Looks good, most importantly the date is earlier. Maybe the following link helps also: https://stackoverflow.com/questions/2979059/rails-is-the-version-number-in-schema-rb-used-for-anything

spec/features/notes_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+
+describe "Notes", type: :feature do
+  before(:each) do
+    @alice = User.new(id: 1, email: "alice@example.com", password: "1234567")
```
chrisma: Want to use the User factory here (`spec/factories/users.rb`)?

spec/features/notes_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    assert @alice.save
+    @bob = User.new(id: 2, email: "bob@example.com", password: "9876543")
+    assert @bob.save
+    login(@alice)
+  end
+
+  # TODO: Split up
+  it "can create a new note" do
+    visit notes_path
+    click_link "New Note"
```
chrisma: Something like `click_link new_note_path` might be less brittle

spec/features/notes_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    fill_in "Title", with: "My Title"
+    fill_in "Content", with: "My Content"
```
chrisma: Maybe the following (using the `build` of a user factory for `new_note`)? 

```Ruby
fill_in "Title", with: new_note.title
fill_in "Content", with: new_note.content
```

spec/features/notes_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    fill_in "User", with: @bob.id
+    click_button "Create Note"
+
+    visit notes_path
+    expect(page).to have_text("My Title")
+    expect(page).to have_text("My Content")
```
chrisma:
```Ruby
    expect(page).to have_text(new_note.title)
    expect(page).to have_text(new_note.content)
```

spec/rails_helper.rb
chrisma: Does this help? https://github.com/heartcombo/devise/wiki/How-To:-sign-in-and-out-a-user-in-Request-type-specs-(specs-tagged-with-type:-:request)

felixauringer: The line is necessary to access the devise test helpers (e.g. `sign_in`) in request specs.

spec/features/notes_spec.rb
rheaSPK: I actually wouldn't split that up since this spec file describes the note controller, and creating a node IS part of the node controller

app/controllers/notes_controller.rb
rheaSPK: This doesn't work like that since just notes know by which user they were created, not the other way around

chrisma: But you could implement it, if you wanted to, right? :) (e.g. by changing the `user` model)
Just a thought

felixauringer: I agree with Christop here, for `current_user.notes` to work one should only have to add `has_many :notes` to the user model (see [here](https://guides.rubyonrails.org/association_basics.html#the-has-many-association)).  
And besides that, `where` also accepts hashes and I think `Note.where(creator_user_id: current_user.id)` looks better than using a string argument.

db/migrate/20201211160500_add_creator_user_to_notes.rb
rheaSPK: @chrisma Hi, right now we just are just saving the user_id of the current_user in the notes model. For that we have the attribute `creator_user_id`. So how can you add the support for `@note.creator = current_user` and why is that necessary?

.idea/.generators
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
LinqLover: Argh! Please remove again :-)

LinqLover:
@rheaSPK Could you provide a short update on your proceedings with this PR? Does it really belong to our board's column "Review" but not rather "Doing" again? Thanks!

app/controllers/notes_controller.rb
rheaSPK: I don't see an issue with that one

.idea/.generators
rheaSPK: Done

db/migrate/20201211160500_add_creator_user_to_notes.rb
rheaSPK: Okay, already done

LinqLover:
Yahoo, our first merged PR! :tada:

PR #41, opened 2020-11-27
Add vagrant setup

sirkrypt0:
This adds all files needed to setup the local environment using Vagrant.

Closes #21

PR #56, opened 2020-12-04
Proposal: Add debugger configuration for VS Code

LinqLover:
I am aware of the convention that editor files are usually added to the `.gitignore`, but in this particular case, I cannot imagine how you should want to configure VS Code else, so I am proposing this patch in order to facilitate all other VS Coders to use the debugger for RoR. What do you think? IMO, this is not more tool-specific than the new Vagrantfile (#41). :-)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -70,4 +70,5 @@ capybara-*.html
 .rspec
 
 # Ignore editor configs
-/.vscode
+/.vscode/*
+!/.vscode/launch.json
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,39 @@
+{
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"name": "RSpec - all",
+			"type": "Ruby",
+			"request": "launch",
+			"program": "${workspaceRoot}/bin/bundle",
+			"args": [
+				"exec", "rspec",
+				"-I",
+				"${workspaceRoot}"
+			]
+		},
+		{
+			"name": "RSpec - active spec file only",
+			"type": "Ruby",
+			"request": "launch",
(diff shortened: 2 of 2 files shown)


felixauringer:
> I am aware of the convention that editor files are usually added to the `.gitignore`, but in this particular case, I cannot imagine how you should want to configure VS Code else, so I am proposing this patch in order to facilitate all other VS Coders to use the debugger for RoR. What do you think? IMO, this is not more tool-specific than the new Vagrantfile (#41). :-)

I would say that editor files and directories should never be added to the `.gitignore`. There are so many editors out there that every project would need a really long `.gitignore` to exclude all of their files. In my opinion `.gitignore`'s main purpose is to exclude files that are created by the application and shouldn't be in the repository, e.g. logs and caches. Editor files belong to the developer and that is also the person that should avoid adding them to the repository. An easy way would be a global `.gitignore` that you can link in your git config. This [blogpost](https://blog.martinhujer.cz/dont-put-idea-vscode-directories-to-projects-gitignore/) explains it in a bit more detail. I'm not sure about the `Vagrantfile` though.

LinqLover:
Hmm, I see your point, but wouldn't it simply provide more convenience? In today's survey, we have seen that a number of persons favor VS Code and we could save them some work by providing the pre-defined configuration file. I would also agree if asked to adding some RubyMine config files if they are not large, high-frequently changed, or pollute the global tree structure of the repository.

LinqLover:
Well agreed.

---

Let's talk about something different: This comment (https://github.com/hpi-swt2/connections-portal/pull/56#issuecomment-738951815) of you @frcroth has now been referenced 4 times in the repository because people pushed their editor configs by accident. Can we do something to prevent further waste of time for this reason?

PR #66, opened 2020-12-10
Created working contact list on contacts page

PatriciaSowa:
Co-authored-by: PatriciaSowa <patricia.sowa@student.hpi.de>

### DIFF
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -1,5 +1,7 @@
 class ContactsController < ApplicationController
+
   def show
-    render params[:page]
+    @contacts = current_user.contacts
+
   end
 end
--- a/app/controllers/contacts_controller.rb
+++ b/app/controllers/contacts_controller.rb
@@ -1,0 +1,5 @@
+class ContactsController < ApplicationController
+  def show
+    render params[:page]
+  end
+end
--- a/app/views/contacts/show.html.erb
+++ b/app/views/contacts/show.html.erb
@@ -1,0 +1,18 @@
+<h1>My Contacts</h1>
+
+<% if user_signed_in? %>
+  <table>
+    <thead>
+    <tr>
+      <th>Email</th>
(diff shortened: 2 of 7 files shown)


app/controllers/contacts_controller.rb
```diff
@@ -0,0 +1,5 @@
+class ContactsController < ApplicationController
+  def show
+    render params[:page]
```
Paula-Kli: I think this is implicit and you can implement an empty show method except you want to show something else

app/views/contacts/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+<% if user_signed_in? %>
+  <table>
+    <thead>
+    <tr>
+      <th>Email</th>
+    </tr>
+    </thead>
+    <tbody>
+    <% current_user.contacts.each do |contact| %>
```
Paula-Kli: I am not sure about it but do you think it is a good way to call the database from a view? I am just thinking about it since we are doing the same thing in der home view?

Paula-Kli (changes requested):

app/controllers/contacts_controller.rb
Paula-Kli: on the other hand: why do you need it?

app/views/contacts/show.html.erb
Paula-Kli: We could maybe in the controller something like: `@contacts = current_user.contacts` in the show action and then we can use @contacts here in the view

Jostafarr (approved):

app/controllers/contacts_controller.rb
SanJSp: was removed

app/views/contacts/show.html.erb
SanJSp: was improved

Paula-Kli (approved):

PR #52, opened 2020-12-04
Add profile page for user

frcroth:

PR #86, opened 2020-12-11
Add social accounts and handles

### LINKED ISSUE #69: [User Story] Store User Handles
# Description
As a user I want to store various handles to each contact.

# Conditions of satisfaction
- [ ] I can store Telegram, Facebook, Twitter, GitHub, GitLab@HPI, Slack Discord handles
- [ ] Even if possible, the handle will redirect to a corresponding page
- [ ] the redirect will open a new tab

### LINKED ISSUE #112: Authentication for social handles
Depends on #86. Make sure no one can change the social handles for another user. Security concern.

Originally proposed here: https://github.com/hpi-swt2/connections-portal/pull/86#issuecomment-756933628

Alexander-Dubrawski:
This PR closes #69 and closes #112.

The goal of this PR is that a user can store various handles to different social networks like GitHub, Facebook, ...
It is implemented the following way: 
* A `SocialAccount` model was added.
* `SocialAccount` has the attributes `social_network` and `user_name`
* The `User` model has a one to many association with the `SocialAccount` model (https://guides.rubyonrails.org/association_basics.html#the-has-many-association). 
* I used nested routes (https://guides.rubyonrails.org/routing.html#nested-resources) between the `User` and `SocialAccount` model. 
* Inside the `social_accounts_helper` (`helper/social_accounts_hlper.rb`) we have a method that returns all supported social networks and has a function that builds the appropriate link.
* We also check if the user is logged in inside the `social account controller`

This Pr only implements the logic and not the style (css). At the moment it looks as following: 

Screenshot 2021-01-09 at 19 52 16

Screenshot 2021-01-09 at 19 52 27

We also display an error message if the values to add a new social network are not valid:

Screenshot 2021-01-09 at 19 57 53

## TODO:
- [x] Add missing e2e tests for editing social accounts 
- [x] Add missing e2e tests for adding a scocial account
- [x] Add Remove functionality
- [x] Add link to social account page
- [x] Add select menu for supported social accounts when adding one
- [x] Format output of accounts (html table)
- [x] Fix typos in model and db
- [x] Fix remaining test/write missing tests
- [x] Add authentication (moved to #112)


frcroth:
#52 is now merged.

LinqLover:
@florian-papsdorf Kind push :)

> Some things are not really clear in the user story but important for the implementation. Like I understand it all the social accounts (or handlers) are displayed on the profile page correct? So the user would for example add a social network like Twitter in his settings and all other users would see a handler for Twitter on his profile page. Is that the desired behavior?

But IMHO the proposed interpretation sounds good. :+1:

app/models/user.rb
```diff
@@ -1,5 +1,8 @@
 # An application user, uses the `devise` library
 class User < ApplicationRecord
+  # Every user can have multiple social accounts like GitHub, Telegram, ...
+  has_many :social_accounts
```
LinqLover: Possible edge case: Do we have any check to make sure that you cannot assign two Twitter handles to one account? Or do we do not want such a limitation at all?

app/views/users/edit.html.erb
```diff
@@ -30,4 +30,17 @@
...
+
+<hr />
+<% if @user.social_accounts.any? %>
+  <h2> Social Accounts </h2>
+  <% @user.social_accounts.each do |social_account| %>
+    <p>
+      <%= render 'social_accounts/social_account_setting', user: @user, social_account: social_account %>
+    </p>
+  <% end %>
+
```
LinqLover:
```suggestion
```

config/routes.rb
```diff
@@ -7,9 +8,11 @@
   root to: 'home#index'
 
   resources :users, only: %i[show edit update] do
+    resources :social_accounts
     member do
       patch 'status', to: 'users#update_status', as: 'update_status'
       patch 'add_contact'
     end
   end
+
```
LinqLover:
```suggestion
```

db/schema.rb
```diff
@@ -21,6 +21,15 @@
     t.index ["user_id"], name: "index_notes_on_user_id"
   end
 
+  create_table "social_accounts", force: :cascade do |t|
```
LinqLover, line 34: Aren't we missing a foreign key definition here? The same change would be required to the migration, too, accordingly.

spec/features/user_profile_spec.rb
```diff
@@ -5,6 +5,10 @@
   let(:user2) { FactoryBot.create :user }
 
   before do
+    user.social_accounts.create(social_network: "GitHub", user_name: "SomeGitUserName")
```
LinqLover: Would it be possible to use factories for these data?

spec/helpers/social_accounts_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+# the SocialAccountsHelper. For example:
+#
+# describe SocialAccountsHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
+RSpec.describe SocialAccountsHelper, type: :helper do
```
LinqLover: TODO

spec/models/social_account_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+    it "is creatable using a factory" do
+        expect(@social_account).to be_valid
+    end
+
+    it "is not valid without an social network" do
+        @social_account.social_network = ""
+        expect(@social_account).to_not be_valid
+    end
+
+    it "is not valid without an user_name" do
```
LinqLover:
```suggestion
    it "is not valid without a user name" do
```

spec/models/social_account_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'rails_helper'
```
LinqLover, line 1: Indentation of this file (and probably others as well) mismatch! Make sure to use 2 spaces for indentation. See Rubocop recommendations. :-)

spec/models/social_account_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+        @social_account = FactoryBot.build(:social_account)
+    end
+
+    it "is creatable using a factory" do
+        expect(@social_account).to be_valid
+    end
+
+    it "is not valid without an social network" do
+        @social_account.social_network = ""
+        expect(@social_account).to_not be_valid
```
LinqLover: (Hm, some specs in our solution are using `not_to`, others are using `to_not`. Should we agree on one solution for the whole repo? Maybe sharpen `RuboCop::Cop::RSpec::NotToNot` in our linter?)

spec/models/user_spec.rb
```diff
@@ -54,4 +54,20 @@
     @user.contacts << contact
     expect(@user.contacts).to include(contact)
   end
+
+  it "should have no relationship to social accounts" do
+      expect(@user.social_accounts).to be_empty
+  end
+
+  it "can have many social accounts" do
+      user_with_social_accounts = User.create(email: "foofoofoo@exmaple.com", password: "super_save_password")
```
LinqLover: Use factories?

spec/requests/social_accounts_request_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+        expect(response).to have_http_status(:success)
+    end
+  end
+
+  describe "POST /users/:user_id/social_accounts" do
+    context "with valid parameters" do
+        it "creates a new social account" do
+            user = FactoryBot.build(:user)
+            user.save
+            post "/users/1/social_accounts", params: { social_account: {social_network: "GitHub", user_name: "Foo" }}
```
LinqLover: Factories?

spec/views/social_accounts/new.html.erb_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe "social_accounts/new.html.erb", type: :view do
```
LinqLover: TODO

LinqLover (commented):
Non-exhaustive list of first impressions. Overall readability is very great!

app/models/user.rb
LinqLover: Our PO liked the absence of this limitation.

spec/models/social_account_spec.rb
LinqLover: Solved via bca5892b05d79b03e617b1d8e77268e7d5dfd114

app/views/users/edit.html.erb
LinqLover: Solved via ad8307df4e6c98931234e64fdf72ebf8cd1b8113

config/routes.rb
LinqLover: Solved via ad8307df4e6c98931234e64fdf72ebf8cd1b8113

spec/models/social_account_spec.rb
LinqLover: Solved via ad8307df4e6c98931234e64fdf72ebf8cd1b8113

frcroth:
Can you add authentication to the controller? While skimming the code, it seems as if everyone could add social accounts for everyone else, if the route is known (Correct me if I'm wrong)

LinqLover:
This might be true, I was not aware of this concept. Thanks for the pointer!

LinqLover:
I found some todo comments on my machine regarding this PR:

- edit without new html page
- "User name" and :user_name should become "Username" and :username

app/views/social_accounts/edit.html.erb
```diff
@@ -0,0 +1,30 @@
+<h1>Edit Social Account</h1>
```
frcroth: I18n

app/views/users/edit.html.erb
```diff
@@ -1,7 +1,18 @@
 <h1>Edit profile</h1>
```
frcroth: I18n

app/views/users/edit.html.erb
```diff
@@ -1,7 +1,18 @@
...
 <hr />
-
+<h2>Profile</h2>
```
frcroth: I18n

app/views/users/edit.html.erb
```diff
@@ -30,4 +41,20 @@
...
     <%= form.submit %>
   </p>
-<% end %>
\ No newline at end of file
+<% end %>
+
+<hr />
+
+<% if @user.social_accounts.any? %>
+  <h2> Social Accounts </h2>
```
frcroth: I18n

app/views/users/edit.html.erb
```diff
@@ -30,4 +41,20 @@
...
+  <% @user.social_accounts.each do |social_account| %>
+    <% if !social_account.new_record? %>
+      <p>
+        <%= render 'social_accounts/social_account_setting', user: @user, social_account: social_account %>
+      </p>
+    <% end %>
+  <% end %>
+<% end %>
+
+<h3>Add a Social Account</h3>
```
frcroth: I18n

app/views/users/show.html.erb
```diff
@@ -23,6 +23,17 @@
   </tr>
 </table>
 
+<% if @user.social_accounts.any? %>
+  <h2> Social Accounts </h2>
```
frcroth: I18n

spec/features/user_profile_spec.rb
felixauringer: Yes! One idea could be to add traits for the different networks

db/schema.rb
felixauringer: The change would only be required in the migration, this file should never be edited manually ;)

config/routes.rb
```diff
@@ -1,4 +1,5 @@
 Rails.application.routes.draw do
+  get 'social_accounts/new'
```
felixauringer: Why is this route necessary? `resources :social_accounts` creates a similar route (but in the user scope)

app/views/users/show.html.erb
felixauringer: +1

app/views/users/edit.html.erb
```diff
@@ -30,4 +41,20 @@
...
+<%= render 'social_accounts/form', social_account: @social_account %>
```
felixauringer: it is not necessary to pass the variable as the partial also can access the instance variable `@social_account`

app/views/social_accounts/new.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>SocialAccounts#new</h1>
```
felixauringer: ?

app/views/social_accounts/_social_account_setting.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render 'social_accounts/social_account', social_account: social_account %>
+<td>
+    <%= link_to 'Edit', edit_user_social_account_path(user, social_account) %>
+    <%= link_to 'Remove', user_social_account_path(user, social_account), method: :delete %>
+</td>
```
felixauringer: I18n

app/views/social_accounts/_social_account.erb
```diff
@@ -0,0 +1,9 @@
+<td>
+	<%= social_account.social_network %>
+</td>
+<td>
+	<%= social_account.user_name %>
+</td>
+<td>
+    <%= link_to 'Link', generate_link(social_account) %>
```
felixauringer: I18n

app/views/social_accounts/_form.html.erb
```diff
@@ -0,0 +1,31 @@
+<%= form_with(model: [@user, social_account], local: true) do |form| %>
```
felixauringer: dry! (see the edit view)

app/views/social_accounts/_form.html.erb
```diff
@@ -0,0 +1,31 @@
...
+  <% if self.controller.is_a?(SocialAccountsController) %>
```
felixauringer: why should a different controller render this? Seems like a hacky workaround

app/models/social_account.rb
```diff
@@ -0,0 +1,4 @@
+class SocialAccount < ApplicationRecord
+    validates :social_network, presence: true, allow_blank: false
```
felixauringer: if an attribute is present, it cannot be blank, so one of these validations is redundant

app/helpers/social_accounts_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+      if network == "Slack"
+        return "https://slack.com/"
+      end
+      if network == "Discord"
+        return "https://discordapp.com/#{social_account.user_name}"
+      end
+      return "#"
+  end
+  def get_supported_social_networks 
+    return [["GitHub", "GitHub"], ["Telegram", "Telegram"], ["Facebook", "Facebook"], ["Twitter", "Twitter"], ["GitLab@HPI", "GitLab@HPI"], ["Slack", "Slack"], ["Discord", "Discord"]]
```
felixauringer: I18n

app/helpers/social_accounts_helper.rb
```diff
@@ -0,0 +1,30 @@
+module SocialAccountsHelper
+  def generate_link(social_account)
+    network = social_account.social_network
```
felixauringer: `case` might be better here

app/controllers/social_accounts_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  def social_account_params
+    params.require(:social_account).permit(:social_network, :user_name)
+  end
+
+  def authorize
+    authenticate_user!
+    if !(current_user.id.to_s == params[:user_id])
+      redirect_to root_path
```
felixauringer: maybe add a flash notice here?

app/controllers/social_accounts_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    @social_account = SocialAccount.find(params[:id])
+  end
+
+  def social_account_params
+    params.require(:social_account).permit(:social_network, :user_name)
+  end
+
+  def authorize
+    authenticate_user!
+    if !(current_user.id.to_s == params[:user_id])
```
felixauringer: `!=` would make the parantheses not necessary

app/controllers/social_accounts_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+    if @social_account.update(social_account_params)
+      redirect_to edit_user_path(@social_account.user_id)
+    else
+      @user = User.find(@social_account.user_id)
+      render :edit
+    end
+  end
+
+  def index
+    # WORKAROUND because URL is wrong after creating an invalid account from
```
felixauringer: so it is not possible to list all social accounts? :sweat_smile:

app/controllers/social_accounts_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+  include SocialAccountsHelper
+  before_action :authorize, :set_social_account, only: [:show, :edit, :update, :destroy]
+  helper_method :generate_link,  :get_supported_social_networks
+
+  def new
+  end
+
+  def edit
+    @user = User.find(params[:user_id])
+    @social_account = @user.social_accounts.find(params[:id])
```
felixauringer: why do you set the social_account explicitly? Shouldn't this be done by `set_social_account`?

app/controllers/users_controller.rb
```diff
@@ -1,16 +1,27 @@
...
+    # prototype for create social account form
+    @social_account = @user.social_accounts.build #
+  end
 
   def update
-    return render :edit unless @user.update(user_params)
-
-    redirect_to @user
+    # prototype for create social account form
+    if @user.update(user_params)
```
felixauringer: this can be simplified to `return redirect_to @user if @user.update(user_params)`

spec/features/user_edit_profile_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'rails_helper'
+
+RSpec.describe "Users profile page", driver: :selenium_headless, type: :feature do
+  let(:user) { FactoryBot.create :user }
+
+  before do
+    user.social_accounts.create(social_network: "GitHub", user_name: "SomeGitUserName")
+    user.save()
```
felixauringer: one save should be enough ;)

spec/features/user_edit_profile_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    expect(page).to have_text("GitHub")
+    expect(page).to have_text("SomeGitUserName")
+    expect(page).to have_text("Telegram")
+    expect(page).to have_text("SomeTelegramUserName")
+  end
+
+  it 'provides link to edit social account' do
+    visit edit_user_path(user)
+    page.all(:link, 'Edit')[0].click
+    expect(find_field("Social network").value).to eq "GitHub"
```
felixauringer: tests should also use I18n for the text of labels, buttons, ...

spec/features/user_edit_profile_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    expect(find_field("User name").value).to eq "SomeGitUserName"
+  end
+
+  it 'provides link to remove social account' do
+    visit edit_user_path(user)
+    
+    page.all(:link, 'Remove')[1].click
```
felixauringer: If I would add another link named 'Remove`, this test would probably break because of the magic index

spec/requests/social_accounts_request_spec.rb
```diff
@@ -0,0 +1,59 @@
+require 'rails_helper'
+
+RSpec.describe "SocialAccounts", driver: :selenium_headless, type: :request do
+  describe "GET /new" do
+    it "returns http success" do
+      get "/social_accounts/new"
```
felixauringer: use rails' path helpers here

spec/requests/social_accounts_request_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+    end
+  end
+
+  describe "POST /users/:user_id/social_accounts" do
+    context "with not valid parameters" do
+      it "creates a new social account" do
+        user = FactoryBot.build(:user)
+        user.save
+        sign_in user
```
felixauringer: a lot of code duplication in here, especially these three lines

felixauringer (changes requested):
You should really use `I18n`! I'm not sure how strict our rubocop is but there are a lot of things that it normally would warn you about, e.g. missing newlines, wrong indentation, ... Just keep in mind to run it before merging :)

rheaSPK (commented):
Basically @felixauringer review said everything I would've said. Please add traits for the user factory instead of creating account handles manually in the tests. If @felixauringer comments are fixed, I'm happy too. Frontend needs a mobile first approach (maybe addable this sprint)
And maybe you could split the PR next time so that it's easier to understand

felixauringer:
> And maybe you could split the PR next time so that it's easier to understand

I would really support this last point :+1:   
There would also be less review comments and it would be easier to incorporate these.

Alexander-Dubrawski:
@felixauringer thanks for all your feedback 🙃 . I will look into it today.

app/helpers/social_accounts_helper.rb
Alexander-Dubrawski: The names of the social networks are in all languages the same 🙃 . So I think we do not need to translate them.

felixauringer: I think that is only true if you already know which languages should be supported in the future. At openHPI we support Chinese and Russian and they do not use the latin alphabet so maybe (I do not know) services could have another name in countries like these. It would then be a lot of work to find all the things previously considered as language independent so in my opinion all strings displayed to a user should be extracted to the locale files.

app/views/social_accounts/_form.html.erb
Alexander-Dubrawski: We are creating a *social account* in the `social account controller` but we are sending the request from the `user controller`. We are doing this since we want to add a social account from the *edit my profile page*. If the account was created unsuccessfully we want to render the user's edit view but still be in the processing of the current request so that the user can correct her/his input. The problem is now that we need to render the error message. That's we are needing this workaround.

Alexander-Dubrawski: I can have a look later if there is a more elegant solution.

app/helpers/social_accounts_helper.rb
Alexander-Dubrawski: I'm personally more a fan of guard clauses 😜 .

felixauringer: But wouldn't you agree that

```ruby
def generate_link(social_account)
  case social_account.network
  when "GitHub"
    "https://github.com/#{social_account.user_name}"
  when "Telegram"
    "https://t.me/#{social_account.user_name}"
  when "Facebook"
    "https://www.facebook.com/#{social_account.user_name}"
  when "Twitter"
    "https://twitter.com/#{social_account.user_name}"
  when "GitLab@HPI"
    "https://twitter.com/#{social_account.user_name}"
  when "Slack"
    "https://slack.com/"
  when "Discord"
    "https://discordapp.com/#{social_account.user_name}"
  else
    '#'
  end
end
```

is much more readable?

app/controllers/social_accounts_controller.rb
Alexander-Dubrawski: Not sure why this is handled inside the index action. This should be handled inside the show action. If we are editing a social account (inside the edit view of the social account) and an error occurs we show the error message so that the user can correct the error. If the user sends manually from his browser a new request (for example by rerendering the page)  we need to handle that case.

app/views/social_accounts/_form.html.erb
felixauringer: Correct me if I am wrong but `<% if social_account.errors.any? %>` ensures that only the relevant errors are displayed here?

spec/models/user_spec.rb
```diff
@@ -55,6 +55,21 @@
     expect(@user.contacts).to include(contact)
   end
 
+  it "should have no relationship to social accounts" do
+    expect(@user.social_accounts).to be_empty
+  end
+
+  it "can have many social accounts" do
+    user_with_social_accounts = FactoryBot.build(:user)
+    user_with_social_accounts.save
```
felixauringer: shouldn't be necessary because you never actually read the database, do you?

Alexander-Dubrawski: We need to save the parent before calling create. But I will change the call on the FactoryBot to create.

spec/features/user_profile_spec.rb
Alexander-Dubrawski: I will have a look at it. But not sure if today.

app/helpers/social_accounts_helper.rb
Alexander-Dubrawski: From my perspective not. But this is something super subjective, so there is not really a right or wrong. @LinqLover how do you see it?

spec/models/user_spec.rb
felixauringer: That's a good idea!

app/helpers/social_accounts_helper.rb
felixauringer: Then it's fine for me :+1:

app/views/social_accounts/_form.html.erb
Alexander-Dubrawski: Ah, I think I fixed it without knowing it. Since `social_account ` is now an instance variable we do not have the problem anymore 🙈 .

spec/models/user_spec.rb
felixauringer: Now that you use `create` this line shouldn't be necessary anymore

app/controllers/social_accounts_controller.rb
Alexander-Dubrawski: I removed the workaround. If we try to reload the page  (inside the browser) while editing the social account, we will stay at the edit page and the index action is doing nothing.

app/controllers/social_accounts_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+      render :edit
+    end
+  end
+
+  def show
+    @user = User.find(@social_account.user_id)
+    render :edit
+  end
+
+  def index
```
felixauringer, line 54: If it is not used I think there should be no action (and route at all). I didn't fully understand why this is needed but it *should* work without it

app/controllers/social_accounts_controller.rb
```diff
@@ -0,0 +1,77 @@
+class SocialAccountsController < ApplicationController
+  include SocialAccountsHelper
+  before_action :authorize, :set_social_account, only: [:show, :edit, :update, :destroy]
+  helper_method :generate_link,  :get_supported_social_networks
+
+  def edit
+    @user = User.find(params[:user_id])
+    set_social_account
```
felixauringer: why do you call `set_social_account` a second time? (see in the `before_action`)

spec/requests/social_accounts_request_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+        post user_social_accounts_path(user), params: { social_account: { social_network: "GitHub", user_name: "Foo" } }
+        expect(response).to redirect_to(edit_user_url(user))
+        expect(user.social_accounts.count).to eq(1)
+      end
+    end
+  end
+
+  describe "POST /users/:user_id/social_accounts" do
+    context "with not valid parameters" do
+      it "creates a new social account" do
```
felixauringer: should it really create a new account? ;)

spec/features/user_edit_profile_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    find_link(href: edit_user_social_account_path(user, social_account1.id)).click
+    # Checks that social account and user name are the default value of the input
+    expect(page.find('#social_account_social_network').value).to eq social_account1.social_network
+    expect(page.find('#social_account_user_name').value).to eq social_account1.user_name
+  end
+
+  it 'provides link to remove social account' do
+    visit edit_user_path(user)  
+    user_social_account_path(user, social_account1.id)
+    find_link("Remove", href: user_social_account_path(user, social_account1.id)).click
```
felixauringer: I18n

config/locales/en.yml
```diff
@@ -63,6 +63,26 @@ en:
     contact_request:
       approved: 'Contact request approved'
       denied: 'Contact request denied'
+    edit:
+      add_social_account_label: 'Create Social Account'
```
felixauringer: I think translations regarding social accounts shouldn't be listed under user (even if they are displayed on the same page)

app/views/users/_flash_messages.html.erb
```diff
@@ -0,0 +1,12 @@
+<div id="error_explanation">
```
felixauringer: Is it really necessary to have a partial included like four times just to display error messages for one model? Shouldn't it be sufficient to have it in the layout?

app/views/home/_flash_messages.html.erb
```diff
@@ -0,0 +1,7 @@
+<div id="error_explanation">
```
felixauringer: same as for the other `_flash_messages` partial

app/views/users/_flash_messages.html.erb
felixauringer: Actually, there already is code to display flash messages in the layout

app/views/social_accounts/edit.html.erb
```diff
@@ -0,0 +1,23 @@
+<h1><%=  I18n.t 'social_accounts.edit.heading' %></h1>
+
+<%= form_with(model: [@user, @social_account], local: true) do |form| %>
```
felixauringer: still not sure why the form partial isn't rendered here

app/controllers/users_controller.rb
```diff
@@ -1,16 +1,25 @@
...
+    @social_account = @user.social_accounts.build #
+  end
 
   def update
-    return render :edit unless @user.update(user_params)
-
-    redirect_to @user
+    # prototype for create social account form
+    return redirect_to @user if @user.update(user_params)
+    handle_error
```
felixauringer, line 20: why not adding normal flash messages here? (That would be displayed by the logic in the layout)

app/controllers/users_controller.rb
```diff
@@ -31,6 +40,13 @@ def add_contact
 
   private
 
+  def handle_error
+      messages = @user.errors.full_messages
+      error_heading = I18n.t 'errors.messages.not_saved.other', count: messages.count, resource: SocialAccount
+      log = {heading: error_heading, messages: messages}
+      flash[:danger] = log
```
felixauringer: I would display the error messages in the form, is there a reason why they are displayed as a flash message?

app/controllers/social_accounts_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+  def destroy
+    @user = User.find(params[:user_id])
+    set_social_account
+    @social_account.destroy
+    redirect_to edit_user_path(@user)
+  end
+
+  private
+
+  def handle_error
```
felixauringer, line 77: same remarks as in the users controller (and not very dry too)

felixauringer (commented):
You really improved a lot of things since my last review :tada:  
Because my remaining threads are mainly code style issues, I would be ok if you do not resolve all my issues before merging (as long as all tests ensure the functionality) and someone resolves them later on. But I think the flash messages and the wrong route should be fixed.

spec/requests/social_accounts_request_spec.rb
Alexander-Dubrawski: Why not? This is the request sent from the form that adds the social account to the user.

app/views/users/_flash_messages.html.erb
Alexander-Dubrawski: I can't find the template in layouts. But I will move the flash message handling there and try to make it more generic.

app/controllers/users_controller.rb
Alexander-Dubrawski: My assumption was to display a heading and then all error messages as bullet points. 

Screenshot 2021-01-15 at 10 53 06

That's why the flash message partial expects a dictionary in form of  `{heading: "Some Heading" , messages: [messageOne, ...] }` . Do you think that it makes sense to display a heading?

app/controllers/users_controller.rb
Alexander-Dubrawski: I assume that all error messages consists of a heading and all messages that belong to the error. It makes things a little bit easier and more generic to display all parts of the error message as a flash message.

app/views/social_accounts/edit.html.erb
Alexander-Dubrawski: Good point.

app/controllers/social_accounts_controller.rb
Alexander-Dubrawski: I agree that this solution is still hacky. I will have a deeper look at it today.

spec/requests/social_accounts_request_spec.rb
felixauringer: the expectation is `expect(user.social_accounts.count).to eq(0)`

spec/models/user_spec.rb
```diff
@@ -53,9 +55,29 @@
...
+    user_with_social_accounts = FactoryBot.create(:user)
+    user_with_social_accounts.social_accounts.create(social_network: "GitHub", user_name: "SomeGitUserName")
+    user_with_social_accounts.social_accounts.create(social_network: "Telegram", user_name: "SomeTelegramUserName")
+
+    expect(user_with_social_accounts.social_accounts).not_to be_empty
+    expect(user_with_social_accounts.social_accounts[0].social_network).to eq("GitHub")
+    expect(user_with_social_accounts.social_accounts[1].social_network).to eq("Telegram")
+  end
+
+  it "adds user to contact request list of other user" do
```
felixauringer: This test already exists below (in the proper scope). It is then also not necessary to move the `let` in a more general context

app/controllers/users_controller.rb
felixauringer: Ah, I now understand what you are doing, the screenshot helped a lot!  
I've always oriented my error explanation at the [rails guide](https://guides.rubyonrails.org/active_record_validations.html#displaying-validation-errors-in-views).  
Your solution adds a little bit overhead because every controller that is using this needs to implement `handle_errors`. But fine for me. Thank you for clarification!

app/controllers/social_accounts_controller.rb
felixauringer: Feel free to write me if you need help debugging it :)

app/views/users/_flash_messages.html.erb
felixauringer: For the mobile version it is here: https://github.com/hpi-swt2/connections-portal/blob/dev/app/views/layouts/mobile.html.erb  
And for the desktop version it is here: https://github.com/hpi-swt2/connections-portal/blob/dev/app/views/layouts/desktop.html.erb

spec/requests/social_accounts_request_spec.rb
Alexander-Dubrawski: That's true 🙈 . I change the description.

Alexander-Dubrawski:
@felixauringer thank's again for all your help. I learned quite a lot and slowly getting a hang of rails. I tried to adjust the actions inside the social account controller. It seems that we are still needing the `index` and `show` action. I tried to add a comment that explains why (these actions are handling edge cases).  I am still new to rails and I wasn't able to find a cleaner solution, if you have an idea please let me know 🙃 .

app/controllers/social_accounts_controller.rb
```diff
@@ -0,0 +1,101 @@
+# Since we are using nested routes the user should only see the edit view of a social account. All information of the
+# social accounts them self is listed inside the users/edit or users/show views. A request to create a social account
```
LinqLover, line 2:
```suggestion
# social accounts themselves is listed inside the users/edit or users/show views. A request to create a social account
```

app/controllers/users_controller.rb
```diff
@@ -1,16 +1,25 @@
...
 
   def show
     @user = User.find(params[:id])
   end
 
-  def edit; end
+  def edit
+    @user = User.find(params[:id])
+    # prototype for create social account form
+    @social_account = @user.social_accounts.build #
```
LinqLover, line 13:
```suggestion
    @social_account = @user.social_accounts.build
```

app/helpers/social_accounts_helper.rb
```diff
@@ -0,0 +1,28 @@
+module SocialAccountsHelper
+  # rubocop:disable Metrics/CyclomaticComplexity
```
LinqLover: What's the reason for this? IMHO we should either generally decide against this rule or follow it in every case. Otherwise, the code structure will become inconsistent.

app/helpers/social_accounts_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+    [
+      %w[GitHub GitHub],
+      %w[Telegram Telegram],
+      %w[Facebook Facebook],
+      %w[Twitter Twitter],
+      ['GitLab@HPI', 'GitLab@HPI'],
+      %w[Slack Slack],
+      %w[Discord Discord]
+    ]
+  end
```
LinqLover: What is the sense of this structure - which looks quite redundant at first glance? :-)

See my comment above - I think a single hash object should suffice for everything.

app/helpers/social_accounts_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+    return "https://github.com/#{social_account.user_name}" if network == 'GitHub'
+    return "https://t.me/#{social_account.user_name}" if network == 'Telegram'
+    return "https://facebook.com/#{social_account.user_name}" if network == 'Facebook'
+    return "https://twitter.com/#{social_account.user_name}" if network == 'Twitter'
+    return "https://gitlab.hpi.de.com/#{social_account.user_name}" if network == 'GitLab@HPI'
+    return 'https://slack.com/' if network == 'Slack'
+    return "https://discordapp.com/#{social_account.user_name}" if network == 'Discord'
+
+    '#'
+  end
```
LinqLover: Ahm, can you please use a hash to refactor this? :-)

app/helpers/social_accounts_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+    network = social_account.social_network
+    return "https://github.com/#{social_account.user_name}" if network == 'GitHub'
+    return "https://t.me/#{social_account.user_name}" if network == 'Telegram'
+    return "https://facebook.com/#{social_account.user_name}" if network == 'Facebook'
+    return "https://twitter.com/#{social_account.user_name}" if network == 'Twitter'
+    return "https://gitlab.hpi.de.com/#{social_account.user_name}" if network == 'GitLab@HPI'
+    return 'https://slack.com/' if network == 'Slack'
+    return "https://discordapp.com/#{social_account.user_name}" if network == 'Discord'
+
+    '#'
```
LinqLover: Is this defined behavior? Where would we see this hash?

app/views/users/index.html.erb
```diff
@@ -1,5 +1,4 @@
 <h1>Users</h1>
-
```
LinqLover, line 2: Why?

config/locales/devise.en.yml
```diff
@@ -55,6 +55,7 @@ en:
       unlocked: "Your account has been unlocked successfully. Please sign in to continue."
   errors:
     messages:
+      authentication_failed: "Authentication failed"
```
LinqLover, line 58: (minor) Sounds a bit too technical. On the other hand, #120 will outdated this change anyway ... :)

spec/features/user_edit_profile_spec.rb
```diff
@@ -0,0 +1,87 @@
+require 'rails_helper'
+
+RSpec.describe 'Users profile page', driver: :selenium_headless, type: :feature do
```
LinqLover, line 3: This test file is quite long, maybe separate it into some `context`s (e.g. for different routes)?

spec/models/user_spec.rb
```diff
@@ -53,10 +56,24 @@
...
 
-  describe 'contact requests' do
-    let(:contact) { FactoryBot.create :user }
-    let(:request) { FactoryBot.create :user }
+  it 'has no relationship to social accounts' do
+    user_with_no_social_accounts = FactoryBot.build(:user)
+    expect(user_with_no_social_accounts.social_accounts).to be_empty
+  end
+
+  # rubocop:disable RSpec/MultipleExpectations
```
LinqLover: Same as above - I don't really see the sense of overriding rubocop rules locally instead of adjusting them globally. :)

LinqLover (changes requested):
@Alexander-Dubrawski As far as I can tell, this PR looks very good and well documented to me! I only left some minor changes regarding overall code quality.

app/helpers/social_accounts_helper.rb
felixauringer: That's the way to provide Items to a html select, one string is the string that is displayed and one string is the string that is sent to the backend (normally this would be an id or something similar. I'm not sure if there is a better way. Maybe one could write something like

```ruby
%w[GitHub Telegram Facebook Twitter GitLab@HPI Slack Discord].map do |network|
  [network, network]
end
```

This outputs the following in my irb:

```ruby
[["GitHub", "GitHub"], ["Telegram", "Telegram"], ["Facebook", "Facebook"], ["Twitter", "Twitter"], ["GitLab@HPI", "GitLab@HPI"], ["Slack", "Slack"], ["Discord", "Discord"]]
```

LinqLover: Thanks for the explanation. Are the word arrays (`%w`) important for the frontend? Otherwise, your code looks equivalent, so we could eliminate a piece of duplication.

felixauringer: `%w` ist just a ruby shorthand for converting a space separated list of words in an array of strings

LinqLover: Great, so the `map` proposals works

app/helpers/social_accounts_helper.rb
felixauringer: I would strongly advise against turning off whole Cops if we cannot fulfill them in one place. I would rather have some places marked as inconsistent than inconstient code all over our application.

Furthermore the following should be equivalent and RuboCop is not offended by it:

```ruby
LINK_MAP = {
    'GitHub' => ->(name) { "https://github.com/#{name}" },
    'Telegram' => ->(name) { "https://t.me/#{name}" },
    'Facebook' => ->(name) { "https://facebook.com/#{name}" },
    'Twitter' => ->(name) { "https://twitter.com/#{name}" },
    'GitLab@HPI' => ->(name) { "https://gitlab.hpi.de.com/#{name}" },
    'Slack' => proc { 'https://slack.com/' },
    'Discord' => ->(name) { "https://discordapp.com/#{name}" }
  }.freeze

  def generate_link(social_account)
    LINK_MAP[social_account.network]&.call(social_account.user_name)
  end
```
(This returns `nil` if the network could not be found)

felixauringer: Oh and the reason for the offense is that RuboCop does not like so many `if`s in one method :sweat_smile:

felixauringer: You can try the code above by adding it and the following snippet to a local ruby file:

```ruby
class SocialNetwork
  attr_accessor :user_name, :network
end

sn = SocialNetwork.new()
sn.network = 'Slack'
sn.user_name = 'felix.auringer'

puts generate_link(sn)
```

LinqLover: Ah yes, that refactoring was quite what I had proposed :-)

> I would strongly advise against turning off whole Cops if we cannot fulfill them in one place. I would rather have some places marked as inconsistent than inconstient code all over our application.

Agreed, but if we agree on using a linter, IMHO we should also reduce the occasions of manually overridden rules to an absolute minimum. Instead, we should refactor the reported places in code according to the linter output ...

felixauringer: I agree, manually adjusting Cops inline should be one of the last opttions. And increasing the allowed cyclomatic complexity would also be an option but if it is really just one method that cannot be refactored properly I wouldn't relax the rules for all other methods too. But I think we have a possible refactoring here so that it shouldn't be necessary to touch RuboCop at all :tada:

spec/models/user_spec.rb
felixauringer: I would just omit `expect(user_with_social_accounts.social_accounts).not_to be_empty`.  
We allowed two expectations in `.rubocop.yml` and this one is redundant because the array is not empty if the first element matches `GitHub`.

spec/requests/social_accounts_request_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  describe 'POST /users/:user_id/social_accounts' do
+    context 'with valid parameters' do
+      it 'creates a new social account' do
+        post user_social_accounts_path(user), params: { social_account: { social_network: 'GitHub', user_name: 'Foo' } }
+        expect(response).to redirect_to(edit_user_url(user))
+        expect(user.social_accounts.count).to eq(1)
+      end
+
+      it 'creates a new social account with invalid attributes' do
```
felixauringer: still the wrong description, isn't it? I would expect something like `does not create a new social account with invalid attributes`

PR #91, opened 2020-12-18
Add layouts and variants to support mobile devices

felixauringer:


app/views/home/_status.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  <div class="dropdown">
+    <button class="btn btn-secondary dropdown-toggle" type="button" id="dropdownMenuButton" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+      Set status
+    </button>
+    <div class="dropdown-menu" aria-labelledby="dropdownMenuButton">
+      <a class="dropdown-item" href="#">available</a>
+      <a class="dropdown-item" href="#">working</a> <!-- should be done dynamically-->
+    </div>
+  </div>
+</div>
```
felixauringer: Newline is missing

app/controllers/application_controller.rb
```diff
@@ -1,7 +1,25 @@
...
+  def set_variant
+    request.variant = :mobile if mobile?
+  end
+
+  def set_layout
+    mobile? ? 'mobile' : 'desktop'
+  end
+
+  def mobile?
+    request.user_agent =~ /Mobile|webOS/
```
felixauringer, line 23: Maybe there is a better way / a more complete regex to distinguish mobile from desktop browsers.

app/views/layouts/_shared_head.html.erb
```diff
@@ -0,0 +1,3 @@
+<title>Connections Portal</title>
+<%= csrf_meta_tags %>
+<%= csp_meta_tag %>
```
felixauringer: Newline is missing

app/views/layouts/mobile.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <%= render "layouts/navbar" %>
+  <h1>Mobile</h1>>
+  <div class="container pt-3">
+    <% if notice %><div class="alert alert-success notice" role="alert"><%= notice %></div><% end %>
+    <% if alert %><div class="alert alert-danger" role="alert"><%= alert %></div><% end %>
+    <%= yield %>
+  </div>
+</body>
+
+</html>
```
felixauringer: Newline missing

felixauringer (commented):
I had a look at all the not mobile related changes

app/controllers/application_controller.rb
felixauringer: It seems like it is bad practice to deliver different html based on the type of browser of the client. See the [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent) for an explanation.

konradh: Instead of serving the variant according to the user-agent string, we could let the user decide which variant they want and only suggest switching to the mobile variant if a mobile browser is detected or the screen has a small resolution.
This could be saved in the session cookie or the mobile variant could be served on a subdomain (the better, but more complicated version, in my opinion).

felixauringer: I think we can leave it like is and switch to a more complicated approach when problems arise.

PR #100, opened 2021-01-07
Feature/search users#23

### LINKED ISSUE #23: Search users by email adress, first name or last name
# Description
As a user, i want to be able to search for other users, using their first, last name or email address

# Hints & Thoughts
* is not dependant to #72 , but both work with mails, devs maybe want to have the same team doing both or choose a convention

# Conditions of satisfaction
- [ ] there is a ui-design wireframe, how it should look like
- [x] There is a new page, reserved to searching for users
- [x] in this page, there is a search-field and a list of users
- [x] if i enter a string in the search-field, the list of users is filtered after my entry (looks for substrings-matching)
- [ ] fix tests

# Extensions
- [ ] I can change the sorting-method of the list
- [ ] i can change how much users the table shows
- [ ] i can navigate between results-pages

antoniusnaumann:
Adds a working search form to search for users by first name, last name, username and email address.

Closes #23 

Bildschirmfoto 2021-01-07 um 16 35 44
Bildschirmfoto 2021-01-07 um 16 37 19


app/controllers/users_controller.rb
```diff
@@ -30,6 +30,21 @@ def add_contact
...
   end
 
+  def search
+     if params[:q].blank?
+        @results = User.all
+     else
+        search_pattern = params[:q].downcase
+        @results = []
+        User.all.each do |user|
+
```
Paula-Kli: is this newline on purpose? It somehow makes it harder to see the `do` block

app/views/layouts/_navbar.html.erb
```diff
@@ -10,7 +10,8 @@
             <li class="nav-item">
                 <%= link_to (fa_icon 'sticky-note', text: 'Notes'), notes_path, class: 'nav-link' %>
             </li>
-            <li class="nav-item"> <%= link_to (fa_icon 'users', text: "My Contacts"), "/contacts", class: 'nav-link' %> </li>
+          <li class="nav-item"> <%= link_to (fa_icon 'users', text: "My Contacts"), "/contacts", class: 'nav-link' %> </li>
+           <li class="nav-item"> <%= link_to (fa_icon 'search', text: "Search"), "/users/search", class: 'nav-link' %> </li>
```
Paula-Kli: why is this an extra tab in the navbar? I thought it would be a field in the normal home/index?

Paula-Kli: for me this extra view does not provide so much more information on the first thought

spec/features/user_search_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'rails_helper'
+
+describe "Filter users by firstname", type: :feature do
+
+     before(:each) do
+        @user1 = User.new(firstname:"Max", lastname:"Mustermann", username:"Maxi", email: "example@example.com")
```
Paula-Kli: wouldn't be `@user1 = FactoryBot.create :user` be sufficient too? Or is there a reason you didn't use the factoryBot?

spec/features/user_search_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        @user2 = User.new(firstname:"Grace", lastname:"Hopper", username:"Profi", email: "demo@demo.com")
+        sign_in @user1
+      end
+
+    it "When users visit the user search page, then it should be possible to filter by firstname" do
```
Paula-Kli: I think a good name would be something like: `it "should be possible to filter by firstname"`

spec/features/user_search_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        visit users_search_path(firstname: "Ma")
+        expect(page).to have_text("Max Mustermann")
+        expect(page).not_to have_text("Grace")
+    end
+
+    it "When users visit the user search page, then it should be possible to filter by lastname" do
```
Paula-Kli: same for that test describtion

Paula-Kli (changes requested):

Paula-Kli:
There are 4 failing tests? Why are they failing?

spec/features/user_search_spec.rb
PatriciaSowa: Yes there is, because it creates random names, which is not very good testable

app/views/layouts/_navbar.html.erb
PatriciaSowa: it was specified like this in the User Story

spec/features/user_search_spec.rb
Paula-Kli: but you can test `@user1.firstname ` ?

spec/features/user_search_spec.rb
Paula-Kli: it should form a sentence if I understood that correctly between `it` and `do`

app/controllers/contact_requests_controller.rb
```diff
@@ -1,26 +0,0 @@
-class ContactRequestsController < ApplicationController
```
Paula-Kli: why is that one deleted?

felixauringer (changes requested):
Why does this remove all the contact request related code? The branch and PR name suggest otherwise but even if this is intended it should be extracted in another PR so this PR is easier to review.

spec/features/user_search_spec.rb
PatriciaSowa: Yes, but what what is a working search_pattern to find this user without enterin his whole name?

felixauringer: You could, for example, only use the first 5 characters of the name with `@user1[0..4]`. IIRC the names are always 26 characters long so there should always be 5 characters in the beginning.

PatriciaSowa: But we also test not to have specific text, so how we can exclude that it is randomly on the page?

felixauringer: Hm, that's a good point. But it should be sufficient to only set the firstname, shouldn't it? That would already make the line more readable.

app/controllers/users_controller.rb
PatriciaSowa: This code doesn't exist anymore

spec/features/user_search_spec.rb
felixauringer: Hm, that's a good point. Maybe you could add traits to the user factory to specify the attributes needed for your test? Or a nested Factory that only defines these 4 attributes.

spec/features/user_search_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    end
+
+    it "should be possible to filter by lastname" do
+       visit users_search_path(lastname: "mann")
+       expect(page).to have_text("Max Mustermann")
+       expect(page).not_to have_text("Grace")
+    end
+
+    it "should be possible to filter by username" do
+        visit users_search_path(user: "Ma")
```
felixauringer: Because the firstname also starts with `Ma` it cannot be assured that it's really the username that is used for filtering

spec/features/user_search_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        expect(page).to have_text("Max Mustermann")
+        expect(page).not_to have_text("Grace")
+    end
+
+    it "should be possible to filter by email" do
+        visit users_search_path(email: "example")
+        expect(page).to have_text("Max Mustermann")
+        expect(page).not_to have_text("Grace")
+    end
+end
```
felixauringer: missing newline ;)

app/controllers/contact_requests_controller.rb
PatriciaSowa: Because we reverted our merge.

felixauringer: But why was is reverted? And even more important, why in this PR and not in a dedicated one?

spec/features/user_search_spec.rb
PatriciaSowa: Right

spec/features/user_search_spec.rb
PatriciaSowa: Thanks

.idea/workspace.xml
```diff
@@ -0,0 +1,285 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
felixauringer: remove

felixauringer:
In my opinion not really reviewable until the reversal is removed or extracted into another PR (depending on what the intention is) :sweat_smile:

spec/features/user_search_spec.rb
Paula-Kli: we are now doing it this way with factorybot and `user[0..4]`

.idea/workspace.xml
Paula-Kli: I think this still needs to be removed from the branch

felixauringer (approved):

Paula-Kli (approved):

spec/features/user_search_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    sign_in user
+  end
+
+  it 'is possible to filter by firstname' do
+    visit search_users_path(search: second_user.firstname[0..2])
+    expect(page).to have_text(second_user.firstname)
+  end
+
+  it 'is possible to filter by lastname' do
+    visit search_users_path(search: second_user.lastname[2..4])
```
PatriciaSowa: Why don't we use [0..2] consistently?

spec/features/user_search_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  it 'does not show non-matching users when filtering by username' do
+    visit search_users_path(search: user.username[0..2])
+    expect(page).not_to have_text(second_user.username)
+  end
+
+  it 'does not show non-matching users when filtering by email' do
+    visit search_users_path(search: user.email[0..2])
+    expect(page).not_to have_text(second_user.email[0..2])
```
PatriciaSowa: That should be the whole email here. 
It could be that the beginnig of two emails is the same and one matches the search pattern and the other doesn't.

PatriciaSowa (changes requested):

spec/features/user_search_spec.rb
felixauringer: I would agree that these tests are not optimal. Because the factory bot uses random names it will happen that both users are found when searching so there will be a failing test from time to time.

PatriciaSowa (approved):

PR #103, opened 2021-01-08
Create activities

mpass99:
The text area in the dashboard can now be used to submit activities to your feed (but there are not displayed by now)


spec/factories/activities.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :activity do
+    content { "MyText" }
+    author { "" }
```
felixauringer: I would add a string here.

config/routes.rb
```diff
@@ -12,4 +12,7 @@
       patch 'add_contact'
     end
   end
+
+  resources :activity, only: :create
```
felixauringer: this should by plural, otherwise the path helper's method names differ from the conventions

spec/requests/activity_request_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe "Activities", type: :request do
+
+  describe "GET /create" do
+    it "returns http success" do
+      get "/activity/create"
```
felixauringer: Normally `create` is a `POST` against `activities`  
see also my comment to `config/routes.rb`

spec/requests/activity_request_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+
+RSpec.describe "Activities", type: :request do
+
+  describe "GET /create" do
```
felixauringer: see below

spec/models/activity_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Activity, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
felixauringer: Maybe it would be nice to test the validations of the model?

spec/requests/activity_request_spec.rb
```diff
@@ -0,0 +1,12 @@
...
+    it "returns http success" do
+      get "/activity/create"
+      expect(response).to have_http_status(:success)
+    end
```
felixauringer, line 21: One could also test that the site is not reloaded

spec/views/activity/create.html.erb_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
```
felixauringer: There is no `create.html.erb`, so this file can be deleted

spec/features/activity_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+RSpec.describe 'Activities', driver: :selenium_headless, type: :feature, js: true do
+  let(:user) { FactoryBot.create :user }
+
+  before do
+    sign_in user
+    visit home_index_path
+  end
+
+  it 'a text box to commit activities' do
```
felixauringer: Thats a strange sentence :D

spec/features/activity_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    expect(page).to have_field('activity-textarea')
+  end
+
+  it 'can be added on the dashboard' do
+    fill_in 'activity-textarea', with: 'This is an activity'
+    find('#add-activity-button').click
+    sleep(0.5) # async
```
felixauringer: Is this still necessary after activating javascript?

app/views/home/_create_activity.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= form_with(url: activity_index_path, method: :post) do |f| %>
```
felixauringer: why don't we use a model for the activity here?

felixauringer: this would look something like `<%= form_with(model: Activity.new) do |f| %>`

app/views/home/_create_activity.html.erb
```diff
@@ -0,0 +1,12 @@
...
+    <div class="card form-group">
+      <label for="activity-textarea">Post new activity</label>
+      <textarea  class="form-control" id="activity-textarea" name="activity[content]" rows="3" placeholder="Just start typing here..."></textarea>
+      <div style="display:none">
+        <%= f.text_field :"activity[user]" , value: current_user.id %>
```
felixauringer: `current_user.id` can be placed in the controller, no extra parameter needed

app/controllers/activity_controller.rb
```diff
@@ -0,0 +1,17 @@
+class ActivityController < ApplicationController
+  # POST /activity
+  def create
+    @activity = Activity.new({content: activity_params[:content], user: User.find(activity_params[:user])})
```
felixauringer: this would benfit from my suggestions from `_create_activity.html.erb`

app/controllers/activity_controller.rb
```diff
@@ -0,0 +1,17 @@
+class ActivityController < ApplicationController
```
felixauringer: The controller name for a resource should always be plural

app/controllers/activity_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  # POST /activity
+  def create
+    @activity = Activity.new({content: activity_params[:content], user: User.find(activity_params[:user])})
+
+    unless @activity.save
+      # TODO: Error Handling
```
felixauringer: I think it would be ok to reload the page to show an error (because this should be an exception). That looks something like this:  
`redirect_to root_path, notice: I18n.t('confirmation.resource_creation', resource: Activity)`

felixauringer (changes requested):
You should run RuboCop on your changes ;)

app/views/home/_create_activity.html.erb
frcroth: What do you mean in the controller?

spec/features/activity_spec.rb
frcroth: Doesn't seem to be

app/views/home/_create_activity.html.erb
frcroth: Doesn't seem to be that simple

felixauringer (approved):

PR #107, opened 2021-01-08
Fixed bug that you're being able to add yourself as contact on index page

### LINKED ISSUE #104: Add yourself
# Description
As a user I don't want to be able to add myself as a contac because it does not have any meaning

SanJSp:
We have simply removed the displayment of the current_user in the table where you can add contacts.
Fixes #104 
Please review.


PatriciaSowa (approved):

app/views/home/index.html.erb
```diff
@@ -25,10 +25,12 @@
       </thead>
       <tbody>
       <% User.all.each do |user| %>
```
felixauringer: The database shouldn't be queried from a view. So one could set `@users = User.where.not(id: current_user.id)` in the index action of the controller and would not need an if here then.

Paula-Kli: we applied it

Jostafarr (approved):

spec/views/users/index.html.erb_spec.rb
```diff
@@ -20,13 +21,17 @@
...
-    it 'has an + button to send contact request' do
+    it 'has a + button to send contact request' do
       expect(rendered).to have_button('+', count: users.length)
     end
 
     it 'does have a + button for not requested user' do
       expect(rendered).to have_css("form.button_to[action=\"/users/#{users.second.id}/contact_requests\"]")
     end
+
+    it 'is not possible to see myself in the list of all users' do
```
felixauringer, line 32: Because this is a view test and you assign only `users` to the instance variable, it is not possible that the current user is displayed on the view because it isn't in the instance variable.
Because a change in the controller should be tested, it would maybe be better to write a feature spec for this.

felixauringer: Because this is a view test and you assign only users to the instance variable, it is not possible that the current user is displayed on the view because it isn't in the instance variable.
Because a change in the controller should be tested, it would maybe be better to write a feature spec for this.

felixauringer (changes requested):

spec/views/users/index.html.erb_spec.rb
```diff
@@ -20,13 +21,17 @@
...
+      expect(rendered).not_to have_text(users.first.email, count: users.length)
+    end
```
PatriciaSowa: That should be also the signed_in_user instead of users.first, shouln't it?

PatriciaSowa (changes requested):

spec/views/users/index.html.erb_spec.rb
felixauringer: I think there are also some other places in this file where it has to be changed

spec/views/users/index.html.erb_spec.rb
Paula-Kli: do you think there should be more covered than is at the moment?

PatriciaSowa (approved):

spec/features/user_index_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  let(:users) { FactoryBot.create_list(:user, 3) }
+  let(:user) { FactoryBot.create :user }
+
+  before do
+    users
+    sign_in user
+    visit users_path
+  end
+
+  it 'is possible to add all users except one' do
```
felixauringer: The description does not match the test case

Paula-Kli: it does since we have all users and the `user` so seeing all users would be users.length+1 but fair it is not a good naming

spec/views/users/index.html.erb_spec.rb
felixauringer: I think everything important is covered but maybe with the wrong test type. I would expect a feature test instead of the helper method `users_to_send_requests_to`. But fine for now, can also be refactored later.

felixauringer (approved):

PR #118, opened 2021-01-12
Multiple contact requests forbidden

### LINKED ISSUE #105: Add contacts multiple times
# Description
As a user I don't want to be able to add one person multiple times as a contact and don't send requests twice

Paula-Kli:
One is not able to send multiple contact requests, if one was send already.

The `+` is not there for those users you have already requested or which are already your contacts.

closes #105


spec/views/users/index.html.erb_spec.rb
```diff
@@ -4,17 +4,29 @@
...
-  it "has an + button to add a contact" do
+  it "has an + button to send contact request" do
+    render
     expect(rendered).to have_button('+', count: @users.length)
   end
+
+  it "does not have a + button for already requested user" do
+    @users.second.contact_requests << @users.first
+    render
+    expect(rendered).not_to have_css('form.button_to[action="/users/2/contact_requests"]')
```
felixauringer: I think there should also be a test for the 'positive' case so that it is clear that the css selector does the right thing. Otherwise one could use any css selector here and all the tests would be green.

spec/views/users/index.html.erb_spec.rb
```diff
@@ -4,17 +4,29 @@
...
+  end
+
+  it "does not have a + button for already existing contacts" do
+    @users.first.contacts << @users.second
+    render
+    expect(rendered).not_to have_css('form.button_to[action="/users/2/contact_requests"]')
```
felixauringer: One could use the path helpers here (and in the test case above)  
`user_contact_requests GET    /users/:user_id/contact_requests(.:format)                                               contact_requests#index`

app/views/users/index.html.erb
```diff
@@ -10,7 +10,9 @@
   <% @users.each do |user| %>
     <tr>
       <td><%= user.email %></td>
-      <td><%= button_to '+', user_contact_requests_path(user), method: :post, id: user.id %></td>
+      <% unless current_user.contacts.include? user or user.contact_requests.include? current_user %>
```
felixauringer: I think complex logic should rather be in the model than in the view.

Paula-Kli: but in the model we have no idea which user we are currently looking at to not show the plus button

felixauringer: But it is possible to add a method to the model that checks whether another user (given as argument) is either in the contacts or if this user already received a contact request. This would simplify this method call to something like `current_user.<good method name I currently don't know>? user`. This would make it easier to write tests too.

spec/views/users/index.html.erb_spec.rb
Paula-Kli: the path helpers are not available in a view test as far as I tried and found stuff

felixauringer: Ok, then it's fine :+1:

spec/views/users/index.html.erb_spec.rb
Paula-Kli: I added a test - Is that what you intended?

app/views/users/index.html.erb
Paula-Kli: were the changes the ones you intended?
- added a method called `sent_contact_request` - this one checks whether I sent the specified user a contact request before, therefore being in their contact requests myself or whether the specified user is already in my contacts

app/models/user.rb
```diff
@@ -32,6 +32,10 @@ def init
     self.username ||= email.split('@', 2)[0]
   end
 
+  def sent_contact_request(user)
```
felixauringer: The name does not really match the function but fine for me if you add a `?` at the end of the method name because it returns a boolean

app/views/users/index.html.erb
```diff
@@ -10,7 +10,9 @@
   <% @users.each do |user| %>
     <tr>
       <td><%= user.email %></td>
-      <td><%= button_to '+', user_contact_requests_path(user), method: :post, id: user.id %></td>
+      <% if not current_user.sent_contact_request(user) %>
```
felixauringer: If you only use the method this way, why didn't you add a method like `no_contact_to?`? The `not` would not be necessary then (and besides, RuboCop would prefer `!` instead of `not`)

app/views/users/index.html.erb
```diff
@@ -10,7 +10,9 @@
...
+      <td><%=  button_to '+', user_contact_requests_path(user), method: :post, id: user.id  %></td>
```
felixauringer: This line should be indented

spec/views/users/index.html.erb_spec.rb
```diff
@@ -4,17 +4,34 @@
   before(:each) do
     @users = assign(:users, FactoryBot.create_list(:user, 3))
```
felixauringer: This would be a lot more readable if you use `context` to group the tests

felixauringer (changes requested):

app/views/users/index.html.erb
felixauringer: yes, that was what I intended

app/views/users/index.html.erb
Paula-Kli: it is not the latest version but thanks for the advice!

Paula-Kli: I don't think `no_contact_to?` would be a good name, since it is not only the persons you have already as contacts but also the ones you only sent a request to

felixauringer: Agreed, naming's always difficult :sweat_smile: The point was that the method could return a boolean that does not have to be inverted.

spec/views/users/index.html.erb_spec.rb
felixauringer: Nice work!

app/models/user.rb
Paula-Kli: done

app/views/users/index.html.erb
felixauringer: A better proposition: use `unless` instead of `if`

Paula-Kli: is in the current code not aplicable anymore

felixauringer (approved):
Looks really good :+1:

app/views/users/index.html.erb
Paula-Kli: it is isn't it?

PatriciaSowa: The indent must be more because of the if statement

felixauringer: Yes, it currently is indented (see the `Files changed` tab and the `outdated` label at this discussion)

PatriciaSowa (approved):

SanJSp (approved):
Good work everybody!!! :champagne:

antoniusnaumann (approved):

PR #119, opened 2021-01-13
Enable more cops, fix the new offenses and add rubocop CI

felixauringer:
This enables all cops except `Style/FrozenStringLiteralComment` and `Bundler/OrderedGems` and fixes the offenses (there are two places where I disable rubocop because the offense is really intended). This also adds a new job to the CI that fails if RuboCop finds any offense.


config/environments/development.rb
```diff
@@ -14,7 +14,7 @@
 
   # Enable/disable caching. By default caching is disabled.
   # Run rails dev:cache to toggle caching.
-  if Rails.root.join('tmp', 'caching-dev.txt').exist?
+  if Rails.root.join('tmp/caching-dev.txt').exist?
```
sirkrypt0, line 17: Is that the correct way? What about Windows Filepaths?

https://stackoverflow.com/questions/47200605/rails-what-is-the-correct-usage-of-rails-root-join-when-pointing-to-tmp-cachin

konradh (approved):

sirkrypt0 (changes requested):
Well done!

config/environments/development.rb
felixauringer: It's the way Rubocop wanted it, but I will have a look

felixauringer: [This discussion](https://github.com/rubocop-hq/rubocop-rails/issues/195) states that it does not make any difference because `Pathname` abstracts from the used OS. So I would stick with the official RuboCop recommendation until someone complains about it ^^

PR #120, opened 2021-01-15
Gated community

### LINKED ISSUE #95: Do not show navbar links to user-specific pages before logging in
Links to notes, My contacts and Home should only viewed after Login.

LinqLover:
Closes #95.


LinqLover:
Sorry for the fine-granular commits on this PR, but what should I do else? I cannot run the e2e tests locally, see #121.

frcroth:
If you use the tests online you can just amend and force push every time, resulting in only one commit but triggering the CI every time

LinqLover:
> If you use the tests online you can just amend and force push every time, resulting in only one commit but triggering the CI every time

Yeah, I know that, but we already talked about the disadvantages of force-pushing ... However, so far I only needed one test commit at all. :-)

app/controllers/notes_controller.rb
```diff
@@ -45,6 +46,15 @@ def destroy
 
   private # Everything below this line is private
 
+  def authorize
+    authenticate_user!
```
felixauringer: This should not be necessary, I think devise redirects if no one is logged in so everything after `authenticate_user!` should only be executed if someone is logged in. But in this case `current_user` is set, so no need to check for `nil`.

app/controllers/users_controller.rb
```diff
@@ -40,10 +45,24 @@ def user_params
...
+      redirect_to users_path, alert: I18n.t('denial.not_found') unless User.exists?(params[:id])
+      update_user
+    end
+    @user = current_user
+  end
+
+  def authorize_to_update
+    if current_user.id.to_s != params[:id]
+      redirect_to users_path, alert: I18n.t('denial.forbidden')
+      return false
```
felixauringer:
```ruby  
def authorize_to_update!
    return redirect_to users_path, alert: I18n.t('denial.forbidden') if current_user.id.to_s != params[:id]
end
```

This interrupts the controller action if the account does not belong to the logged in user but does nothing if it does

felixauringer: This would also simplify the calls from `return unless authorize_to_update` to `authorize_to_update!` (I would use the exclamation mark because devise uses it too and it possibly can stop the controller action with the redirect)

app/controllers/users_controller.rb
```diff
@@ -40,10 +45,24 @@ def user_params
...
     end
+    true
+  end
+
+  def update_user
```
felixauringer, line 85: Because it is only called once, I think a extra method is not necessary but confusing

spec/features/navbar_spec.rb
```diff
@@ -45,4 +45,19 @@ def toggle_profile_dropdown
       page.execute_script('document.getElementById("navbarProfileDropdown").click()')
     end
   end
+
+  describe 'public page' do
+    before do
+      sign_out user
```
felixauringer: You could move the other before block in the describe above and then wouldn't have to logout here

spec/features/navbar_spec.rb
```diff
@@ -45,4 +45,19 @@ def toggle_profile_dropdown
...
+      visit root_path
+    end
+
+    it 'does not contain a link to the notes page' do
```
felixauringer, line 54: why did you decide not to include tests for all the links?

spec/requests/notes_spec.rb
```diff
@@ -19,29 +19,67 @@
       get notes_url
       expect(response).to be_successful
     end
+
+    context 'when not signed in' do
+      before { sign_out user }
```
felixauringer, line 24: again, I wouldn't sign in and sign out again but rather only sign in if it is really necessary

spec/requests/users/users_show_spec.rb
```diff
@@ -3,8 +3,28 @@
 RSpec.describe 'Show user', type: :request do
   let(:user) { FactoryBot.create :user }
 
+  before { sign_in user }
+
   it 'returns http success' do
```
felixauringer, line 8: maybe it would be good to have this in a context to so that it is clearly that the user is logged in

spec/requests/notes_spec.rb
```diff
@@ -19,29 +19,67 @@
       get notes_url
       expect(response).to be_successful
     end
+
```
felixauringer, line 22: Maybe it would be more readable to split this into multiple files (like the user request specs)

app/controllers/users_controller.rb
```diff
@@ -40,10 +45,24 @@ def user_params
 
   def authorize
     authenticate_user!
-    if current_user.id.to_s == params[:id]
-      @user = current_user
-    else
-      redirect_to root_path
+    return redirect_to root_path, alert: I18n.t('denial.unauthorized') if current_user.nil?
+
+    if params.key?(:id)
```
felixauringer: That's a lot of logic just for setting the `@user` for edit and update :sweat_smile:  
I would be fine with deleting this method, using `authenticate_user!` as before_action and maybe adding a method that just creates the instance variable if necessary.

```ruby
def user
  @user ||= User.find(params[:id])
end
```
It is already catched in the application controller if an ActiveRecord model cannot be found, so that we do not have to put this logic into every controller.

spec/requests/application_spec.rb
```diff
@@ -1,8 +1,8 @@
 require 'rails_helper'
 
-RSpec.describe 'Application', type: :request do
-  it 'renders a 404 if active record model is not found' do
-    get user_path(-1)
-    expect(response).to have_http_status :not_found
-  end
-end
+# RSpec.describe 'Application', type: :request do
```
felixauringer, line 3: What's wrong with this?

app/controllers/notes_controller.rb
```diff
@@ -45,6 +46,15 @@ def destroy
...
+    if current_user.nil?
+      redirect_to root_path
+    else
+      @user = current_user
```
felixauringer: setting the user here is not really necessary, isn't it?

felixauringer (changes requested):
The actions `HomeController#dashboard`, `UsersController#index` and `UsersController#add_contact` are not tested for authentication yet. The `ContactsController` and `ContactsRequestsController` are not authenticated at all :sweat_smile: But I'm fine with doing that in another PR.

I do not really like the logging in and out in the tests. I would rather just login if it is necessary.

app/controllers/notes_controller.rb
LinqLover: Fixed via ad7d4868b17a67f37fd9d0889dbdb3ead82765cb

app/controllers/notes_controller.rb
LinqLover: Very good catch. `@user` is never ever accessed from within in the notes controller or any of its views ...

Fixed in 10a30cce52bd6965150c4115f385d2aedd7c3c35.

app/controllers/users_controller.rb
LinqLover: Thanks for the interesting refactoring! Fixed in 23baabcaa15e41617b15492b120bc74e07c47a6b.

app/controllers/users_controller.rb
LinqLover: Originally it was necessary to keep CYCLO low enough, but thanks to your refactoring proposal, I could now save the extra method via 23baabcaa15e41617b15492b120bc74e07c47a6b.

spec/features/navbar_spec.rb
LinqLover: Solved via 45956613fa8b41e0acba53b7eaf43f8ed19f27ff

spec/features/navbar_spec.rb
LinqLover: It was not a functional requirement. Actually, showing the home button would not harm either because the home controller displays the login page unless signed in.

spec/requests/notes_spec.rb
LinqLover: But this variant reduces unnecessary code duplication, so what's the problem with it? Are our tests so slow ATM that we need to optimize them? :-)

spec/requests/users/users_show_spec.rb
LinqLover: I also thought about this, but many contexts already provide the right set up of notes and attributes etc. so this would probably introduce more duplication than a single `sign_in` statement ...

app/controllers/users_controller.rb
LinqLover: > maybe adding a method that just creates the instance variable if necessary.
> 
> ```ruby
> def user
>   @user ||= User.find(params[:id])
> end
> ```

Nah, this smells like a possible cache invalidation problem to me if we are not careful :-)

> It is already catched in the application controller if an ActiveRecord model cannot be found, so that we do not have to put this logic into every controller.

Interesting point, on the other hand, the current solution is more convenient because you are redirected to the user list automatically ... What's more important for us, simple code or convenient error handling?

spec/requests/application_spec.rb
LinqLover: There is no longer any scenario in which the 404 page is shown because the relevant controllers show custom warnings instead. See https://github.com/hpi-swt2/connections-portal/pull/120#discussion_r559004973

app/controllers/users_controller.rb
felixauringer: IIRC an instance of a Controller is created for every request and deleted after that so caching shouldn't be a problem. Or did you had another case in mind?

> What's more important for us, simple code or convenient error handling?

That's a hard choice, I'm currently thinking if it is possible to parameterize the path but did not have an idea yet. So it's fine leaving it as it is :+1:

spec/requests/application_spec.rb
felixauringer: Maybe the logic could then be tested with another controller that does not have custom warnings yet (e.g. the `NotesController`)

spec/requests/notes_spec.rb
LinqLover: But again, this would require us to duplicate parts of the set-up logic which I'm not a friend of ... Or would it be a common practice in Ruby to try to express the test in multiple partial classes (I have no idea how this would be possible at all)?

spec/requests/application_spec.rb
LinqLover: The `NotesController` redirects unauthorized requests.

app/controllers/users_controller.rb
LinqLover: > Or did you had another case in mind?

Not concretely, was only thinking of the rather complex solution in #86

LinqLover:
@felixauringer Thanks for the detailed review! I have patched as many places in the code as possible.

> The actions `HomeController#dashboard`, `UsersController#index` and `UsersController#add_contact` are not tested for authentication yet. The `ContactsController` and `ContactsRequestsController` are not authenticated at all 😅 But I'm fine with doing that in another PR.

Good point. I have added authentication/authorization for `ContactsController` and `ContactsRequestsController` (via e2e43d8) and added authentication tests to the `HomeController` test (e5490d7) as well as a new test for `UserController#index` test (5d5421a). Only `UsersController#add_contact` remains untested but since there is not any request spec for this controller at all, I would like to defer this indeed.

app/controllers/contacts_controller.rb
```diff
@@ -1,5 +1,18 @@
...
+
+  private
+
+  def authorize!
+    authenticate_user!
+
+    return current_user if current_user.id.to_s == params[:user_id]
+
+    redirect_to users_path, alert: I18n.t('denial.forbidden')
+    nil
```
felixauringer, line 16:
```suggestion
    redirect_to users_path, alert: I18n.t('denial.forbidden') unless current_user.id.to_s == params[:user_id]
```

app/controllers/users_controller.rb
```diff
@@ -40,10 +43,20 @@ def user_params
...
+
+    redirect_to users_path, alert: I18n.t('denial.not_found') unless User.exists?(params[:id])
+    update_user
+  end
+
+  def authorize_to_update!
+    return @user if current_user.id.to_s == params[:id]
+
+    redirect_to users_path, alert: I18n.t('denial.forbidden')
+    nil
```
felixauringer:
```suggestion
    redirect_to users_path, alert: I18n.t('denial.forbidden') unless current_user.id.to_s == params[:id]
```

app/controllers/contacts_controller.rb
felixauringer: Ah I see why you had to change this. I think it would be possible to use `return if authorize!` after my change because it only returns a value when redirecting but then the method name wouldn't be very intuitive. But because that's also not very pretty it' would be ok for me to leave it like it is now :)

spec/requests/home_spec.rb
```diff
@@ -9,13 +9,14 @@
         sign_in user
         get root_path
         expect(response).to have_http_status(:success)
+        expect(response.body).to include('Last Activities')
```
felixauringer: you should use I18n here :)

spec/requests/notes_spec.rb
felixauringer: It's just not necessary and I think tests should as atomar as possible.  
If a commit destroys the login system, all these specs would fail too and it would be harder to find the real problem

spec/requests/users/users_show_spec.rb
felixauringer: You're right, a little bit more code for a better understanding of the tests, but this is so small it should be understandable like this too :+1:

spec/requests/notes_spec.rb
felixauringer: Maybe that's possible but it sounds like a little bit too much overhead :sweat_smile: 
Lucky for you that Christoph 'fixed' this issue by adding

```ruby

# Ignore "Block has too many lines" for specs, `RSpec.describe` blocks can be long
Metrics/BlockLength:
  Exclude:
    - 'spec/**/*.rb'
```

to the rubocop config :D

spec/requests/application_spec.rb
felixauringer: If you removed all usages of this logic, I think the logic itself and the tests for it could be removed too. But I still don't really like to build this ourself in every controller :sweat_smile:

LinqLover: But it allows us to provide more convenient behavior. I would not like to remove this convenience again ...

But I would also not delete the logic because it does no harm but will provide a pretty good fallback for any new controller that does not implement error handling itself.

app/controllers/users_controller.rb
LinqLover: But in this case the return value will not be non-nil if `current_user.id.to_s == params[:id]` evaluates to `true`, will it?

spec/requests/home_spec.rb
LinqLover: Interesting, I was not aware that we should make the tests international as well. Done via e19737e247b450b00059b2c9801d6a78e0f14ab7 :-)

LinqLover:
@felixauringer Could you please react to the latest open question (https://github.com/hpi-swt2/connections-portal/pull/120#discussion_r562678929)? Otherwise, this PR is ready from my side!

spec/requests/application_spec.rb
felixauringer: I agree that this logic provides convenience but after you removed all usages of it, it is pretty much useless. But fine for me.

app/controllers/users_controller.rb
felixauringer: I think that it could work
```ruby
'test' unless 1 == 1
=> nil
'test' unless 1 == 2
=> "test"
```
EDIT: but one would have to use `if` instead of `unless` when calling the method (I think)

LinqLover: If `current_user.id.to_s == params[:id]` evaluates to `true`, the current implementation would return `@user` (not `nil`). But with your change, it would return `nil`, wouldn't it? In the other case, `redirect_to` will return a non-`nil` value instead of `nil`, so the logic is the opposite. Should it be this instead?

```rb
not (redirect_to users_path, alert: I18n.t('denial.forbidden') unless current_user.id.to_s == params[:id])
```

felixauringer: That would be the better solution, but RuboCop will require `!` instead of `not`

LinqLover: Ah, nevermind ... this diff has been outdated anyway.

LinqLover:
All discussions resolved! Can I get an approval? :-)

felixauringer (approved):

LinqLover:
Thanks for the detailed review!

PR #122, opened 2021-01-15
Feature/user filter status

### LINKED ISSUE #31: Filter by current status
## User Story

**As a** user -  
**I want to** filter accounts by their current status -  
**In order to** communicate with users that might currently be willing to do so.

## References
- #68 (non-blocking)

## Conditions of satisfaction
- [x] a list of users who have the status `free for chat` set is displayed
  - [x] it should include 30 entries if available (scroll on overflow)
  - [x] at least first name and last name (username as fallback) and a placeholder for a profile-picture is displayed 
  - [x] the list should be in randomised order

## Implementation Note

Please discuss with the PO if you can think of any more information that might make sense to be included in the account list.

konradh:
Fixes #31 

Besides the list of users with status `nice_to_meet_you` on the dashboard, this adds a `display_name` to the User model that consists of first and last name (if given) and defaults to the username.
For each status of a user, there is a class method in the User class that returns the status string. These methods follow the format `User.status_#{status}`.


felixauringer (approved):

app/controllers/home_controller.rb
```diff
@@ -1,10 +1,15 @@
...
+      filter_users_status
       render 'dashboard'
     else
       render 'index'
     end
   end
+
+  def filter_users_status
+    @users = User.where.not(id: current_user.id).with_status(User.status_nice_to_meet_you)
+    @users = @users.sample(30)
```
frcroth: Magic number

app/controllers/home_controller.rb
```diff
@@ -1,10 +1,15 @@
...
     if user_signed_in?
+      filter_users_status
       render 'dashboard'
     else
       render 'index'
     end
   end
+
+  def filter_users_status
+    @users = User.where.not(id: current_user.id).with_status(User.status_nice_to_meet_you)
```
frcroth: Use self.filter_status

spec/features/dashboard/user_list_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  let(:working_user) { FactoryBot.create :user, current_status: User.status_working }
+
+  before { sign_in user }
+
+  context 'with some users' do
+    let!(:users) { FactoryBot.create_list :user, 5, current_status: User.filter_status }
+
+    before { visit root_path }
+
+    it 'does show the right number of users' do
```
frcroth:
```suggestion
    it 'shows the right number of users' do
```

spec/models/user_spec.rb
```diff
@@ -89,4 +89,37 @@
...
+
+    it 'contains the first name if present' do
+      user = FactoryBot.build(:user, firstname: 'Hasso', lastname: '', username: 'hasso01')
+      expect(user.display_name).to eq('Hasso')
+    end
+
+    it 'contains the username if neither first nor lastname are present' do
+      user = FactoryBot.build(:user, firstname: '', lastname: '', username: 'hasso01')
+      expect(user.display_name).to eq('hasso01')
+    end
```
frcroth, line 123: Does it not include the user name when first and lastname are present?

frcroth (changes requested):

app/controllers/home_controller.rb
felixauringer: Should also be changed in the test

spec/models/user_spec.rb
```diff
@@ -89,4 +89,37 @@
       expect(user.sent_contact_request?(contact)).to be true
     end
   end
+
+  describe 'status scope' do
+    let(:user1) { FactoryBot.create :user, current_status: described_class.status_working }
+    let(:user2) { FactoryBot.create :user, current_status: described_class.status_free_for_chat }
+
+    it 'shows only users with the given status' do
```
frcroth: should not use operator "shows"

frcroth (approved):

PR #131, opened 2021-01-18
#77 Add functionality of the global chat via websocket

### LINKED ISSUE #77: Global chat
## User Story

**As a** user -  
**I want to** have a global live-chat -  
**In order to** engage with the other users that are currently active on the site.

## Dependencies
- [x] builds partially on #31 

## Conditions of satisfaction
- [x] messages can be sent
- [x] messages of other users are received in real-time

## Implementation Note

Building a custom chat here is not a requirement.

Maybe have a look at <https://iridakos.com/programming/2019/04/04/creating-chat-application-rails-websockets>

Note: Maybe later one channel per status

sirkrypt0:


config/routes.rb
```diff
@@ -1,4 +1,6 @@
 Rails.application.routes.draw do
+  resources :room_messages
```
felixauringer: Maybe we should only create the routes that are really needed here

db/migrate/20210115135535_add_global_chat.rb
```diff
@@ -0,0 +1,10 @@
+class AddGlobalChat < ActiveRecord::Migration[6.0]
+  def up
+    Room.create name: 'Global Chat', id: Room::GLOBAL_CHAT_ID
```
felixauringer: Because this room exists in all instances of our software we should think of a way to use I18n

config/initializers/warden_hooks.rb
```diff
@@ -0,0 +1,9 @@
+Warden::Manager.after_set_user do |user, auth, opts|
+  scope = opts[:scope]
+  auth.cookies.signed["#{scope}.id"] = user.id
```
felixauringer, line 7: Why do we have to handle cookies ourselves? We have access to the id of the currently logged in user via `current_user.id`.

app/javascript/packs/application.js
```diff
@@ -9,6 +9,8 @@ require("@rails/activestorage").start()
 require("channels")
 require("bootstrap")
 
+//= require jquery
```
felixauringer: why is this a comment?

sirkrypt0: It's special ruby syntax ^^

sirkrypt0: https://github.com/rails/jquery-ujs

db/migrate/20210115135535_add_global_chat.rb
sirkrypt0: That's a good point. However the name shown here is not used to display the name on the dashboard. That is already done using I18n.

config/initializers/warden_hooks.rb
sirkrypt0: I'll just quote the tutorial we followed here.

> When establishing a websocket connection, we don’t have access to the user session but we do have access to the cookies. So, in order to be able to authenticate the user we need to do some devise related stuff first.

It is used in conjunction with `app/channels/application_cable/connection.rb#find_verified_user`

sirkrypt0: I added the above explanation as a comment.

spec/factories/rooms.rb
```diff
@@ -1,5 +1,5 @@
 FactoryBot.define do
   factory :room do
-    name { "MyString" }
+    name { 'MyString' }
```
frcroth: Make it something more useful

sirkrypt0:
Most likely you need to run `rails db:seed` for normal usage and `RAILS_ENV=test rails db:setup` once to get the global chat.

frcroth:
Send on Enter would be nice

app/assets/stylesheets/room_messages.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the RoomMessages controller here.
```
felixauringer: Could be deleted?

app/assets/stylesheets/rooms.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the Rooms controller here.
```
felixauringer: Could be deleted?

app/controllers/rooms_controller.rb
```diff
@@ -0,0 +1,2 @@
+class RoomsController < ApplicationController
```
felixauringer: Could be deleted? There is no route to this.

app/helpers/room_messages_helper.rb
```diff
@@ -0,0 +1,2 @@
+module RoomMessagesHelper
```
felixauringer: Could be deleted?

app/helpers/rooms_helper.rb
```diff
@@ -0,0 +1,2 @@
+module RoomsHelper
```
felixauringer: Could be deleted?

app/controllers/home_controller.rb
```diff
@@ -2,6 +2,7 @@ class HomeController < ApplicationController
   def dashboard
     if user_signed_in?
       filter_users_status
+      @room_messages = Room.find(Room::GLOBAL_CHAT_ID).room_messages
```
felixauringer: Why don't you use the defined scope (room model) here?

config/initializers/warden_hooks.rb
felixauringer: Thanks for the explanation :+1:

db/migrate/20201211160500_add_creator_user_to_notes.rb
```diff
@@ -1,6 +1,7 @@
 class AddCreatorUserToNotes < ActiveRecord::Migration[6.0]
```
felixauringer, line 1: It may be problematic to change a migration that is already rolled out in production (because it would be rolled back with a `change` method different to the method it was migrated with). But I never liked this broken migration so let's try out what happens :D

spec/models/room_message_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+  it 'is creatable using a factory' do
+    expect(room_message).to be_valid
+  end
+
+  it 'is invalid without a message' do
+    room_message.message = ''
+    expect(room_message).not_to be_valid
+  end
+
+  it 'json should contain the display name key' do
```
felixauringer, line 17: Somehow unnecessary because you try to access this key in the next test, so that would fail in case the key does not exist. But fine to leave.

spec/models/room_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'rails_helper'
+
+RSpec.describe Room, type: :model do
```
felixauringer, line 3: You should test the scope here too

spec/views/home/dashboard.html.erb_spec.rb
```diff
@@ -2,8 +2,12 @@
 
 RSpec.describe 'home/dashboard', type: :view do
   let(:users) { FactoryBot.create_list :user, 3, current_status: User.filter_status }
+  let(:room_messages) { FactoryBot.create_list :room_message, 3, room: Room.find(Room::GLOBAL_CHAT_ID) }
```
felixauringer: Why did you define a scope if you never use it? ^^

app/javascript/channels/room_channel.js
```diff
@@ -0,0 +1,40 @@
+import consumer from './consumer'
+
+function formatDate(date) {
+    return date.getFullYear() + "-" + ("0" + (date.getMonth() + 1)).slice(-2) + "-" + ("0" + date.getDate()).slice(-2) +
```
felixauringer: can't you somehow use the method on the message model?

app/javascript/channels/room_channel.js
```diff
@@ -0,0 +1,40 @@
...
+                    content.find('[data-role="message-author"]').text(data.display_name);
+                    content.find('[data-role="message-author"]').attr('href', `/users/${data.user_id}`);
+                    content.find('[data-role="message-text"]').text(data.message);
+                    content.find('[data-role="message-date"]').text(formatDate(new Date(data.updated_at)));
+
+                    if (data.user_id === document.current_user) {
+                        content.addClass('chat-item-me')
+                    }
+                    $element.append(content);
+                    $element.animate({scrollTop: $element.prop("scrollHeight")}, 1000);
```
felixauringer, line 30: Too much js for me here but the result looks fine :D

felixauringer (changes requested):
A lot of files that could be deleted in my opinion. Very clean PR if you consider that it has over 400 LOC

felixauringer:
And we should somehow test (or ask @chrisma) whether the seed is loaded when the branch is deployed to make sure that the global chat room exists on the deployed system.

spec/views/home/dashboard.html.erb_spec.rb
frcroth: Because we only defined it after everything else

app/javascript/channels/room_channel.js
frcroth: Explain how

sirkrypt0: Technically we could transform the date in the model JSON already, such that the frontend only has to take that and display it. Our thought behind the current version was to keep a unified output from the backend and let the frontend decide on how to display the date.

spec/models/room_spec.rb
felixauringer: Nice work!

app/javascript/channels/room_channel.js
felixauringer: In `spec/features/global_chat_spec.rb` you use the method from the model to check whether the timestamp is displayed correctly and in `views/room_messages/_message.html.erb` you use the method to display the timestamp. So I think you should use the same method when a new message arrives. I think you could use the `as_json` method of the model as already said by Tobias.

felixauringer: You are already doing this for the `display_name` so it should be quite easy ;)

felixauringer: And for maintainability this could also be considered for the link, so one has not so many places to change it if it ever changes :sweat_smile:

app/views/home/_chat.html.erb
```diff
@@ -1,92 +1,32 @@
...
+               message: room_message %>
+  <% end %>
+</div>
+<%= form_with(model: RoomMessage.new, remote: true) do |f| %>
+  <div class="input-group" id="chat-new-message">
+    <%= f.text_area :message, class: 'form-control',
+                    rows: "2", placeholder: I18n.t('chat.placeholder') %>
+    <%= f.hidden_field :room_id, value: Room::GLOBAL_CHAT_ID %>
+    <div class="input-group-append inputWithIcon">
+      <%= f.submit class: "btn btn-primary", id: "post-chat-button", value: I18n.t('chat.send') %>
```
konradh: After the submit button is pressed, the text is changed from `Send` to `Create Room message` for a short moment. Depending on the width of the window, the longer text causes the text field and button to be arranged vertically instead of horizontally. Can this behavior be suppressed?

app/javascript/channels/room_channel.js
```diff
@@ -0,0 +1,40 @@
+import consumer from './consumer'
+
+function formatDate(date) {
+    return date.getFullYear() + "-" + ("0" + (date.getMonth() + 1)).slice(-2) + "-" + ("0" + date.getDate()).slice(-2) +
+        " " + ("0" + date.getHours()).slice(-2) + ":" + ("0" + date.getMinutes()).slice(-2);
+
+}
+
+$(function () {
+    $('[data-channel-subscribe="room"]').each(function (index, element) {
```
konradh, line 4: In some cases, this seems seems to stop working after actions that reload the page without killing the WebSocket connection, like the `redirect_to` when trying to send an empty message, navigating to `My Contacts` and back in the navbar or navigating at the bottom in the mobile variant.
If this happens, one does not receive new messages, but sending still works, because it uses a `POST` request. Also, if this happens, the message container is not scrolled to the bottom.
Only if reloading using `Crtl-r` or similar, that also kills or reloads the ws connection, it works again.

konradh (changes requested):

app/javascript/channels/room_channel.js
sirkrypt0: Fixed in 
8b4fbb8

app/views/home/_chat.html.erb
mpass99: Fixed in Commit 4977e552

sirkrypt0: Fixed in 
4977e55

sirkrypt0:
> Send on Enter would be nice

Added in 
8b4fbb8

app/javascript/channels/room_channel.js
sirkrypt0: Fixed in
f0be623

konradh (approved):

felixauringer (approved):

PR #147, opened 2021-01-22
146 navigation webapp

### LINKED ISSUE #146: Navigation in the web-app
# Description
As a user I want to have an footer and a header to be able to navigate between the different pages of the web-app.  
In the footer there should be 
 - Dashboard
 - Chat
 - Contacts
 - History
 - Settings button

In the Header there should be 
- the App-Title
- a Search
- a Notification button  

# Dependencies

# Conditions of satisfaction
See wireframe at #138

rheaSPK:
Closes #146 
image

Since the dashboard page has it's own footer and doesn't provide independent routes for chatting and dashboard, the current dashboard page looks like this:
image

Note, that chat and dashboard have the exact same routes, that's why the footer shows both be clicked. I expect that the dashboard will be redesigned eventually, that's why I don't see an issue with that right now.
Also there is no real settings page, I've put `edit_user_path` for the current user instead. If there's a settings page eventually, this has to be replaced.
Since we have no history page at all, this link isn't linking anywhere.


LinqLover:
Woah, looks awesome! I don't think we should show two footers at the same page, this would be very unintuitive.

Another hint: Could you use bootstrap-select for the status choice? It's an extenal library but quite simple to use and would look much more better than the current plain HTML solution IMHO. What do you think? :)

rheaSPK:
@LinqLover 
> Woah, looks awesome! I don't think we should show two footers at the same page, this would be very unintuitive.

This is indeed unintuitive, but the whole dashboard has to be redesigned anyway. Right now there is just an example text. I'll ask if someone is already doing this and ask them if they could integrate my solution to their dashboard. The ticket just said it wanted to have a header and a footer, so I guess it editing the dashboard would be out of scope

> Another hint: Could you use bootstrap-select for the status choice? It's an extenal library but quite simple to use and would look much more better than the current plain HTML solution IMHO. What do you think? :)

AFIK I don't use a select dropdown menu in this design because it wasn't in the scoop adding a status choice or anything, so there's no need for bootstrap select. Or what do you mean?

LinqLover:
> > Another hint: Could you use bootstrap-select for the status choice? It's an extenal library but quite simple to use and would look much more better than the current plain HTML solution IMHO. What do you think? :)
> 
> AFIK I don't use a select dropdown menu in this design because it wasn't in the scoop adding a status choice or anything, so there's no need for bootstrap select. Or what do you mean?

Because the dashboard will be redesigned anyway, we leave it as-is for the moment.

app/assets/stylesheets/home_mobile.scss
```diff
@@ -5,7 +5,7 @@
 
 html, body {
   position:relative;
-  height: calc(100% - 28px); /* Subtract the height of the navbar, change if navbar has new design */
+  height: calc(100% - 48px); /* Subtract the height of the navbar, change if navbar has new design */
```
LinqLover, line 8: Maybe use an SCSS constant for all occurrences of `48px`?

spec/features/navbar_spec.rb
```diff
@@ -1,66 +0,0 @@
-require 'rails_helper'
```
LinqLover, line 1: Why did you remove the entire navbar specs?

ubuntu-bionic-18.04-cloudimg-console.log
```diff
@@ -0,0 +1,611 @@
+[    0.000000] Linux version 4.15.0-132-generic (buildd@lgw01-amd64-036) (gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)) #136-Ubuntu SMP Tue Jan 12 14:58:42 UTC 2021 (Ubuntu 4.15.0-132.136-generic 4.15.18)
```
LinqLover, line 1: Please remove again and update your global `.gitignore` :-)

LinqLover (changes requested):
@rheaSPK Only a few formal things, UX is fine :-)

spec/features/navbar_spec.rb
rheaSPK: Because the whole navbar changed (non of the tests are relevant for the new navbar) and I wouldn't test the frontend functionality

rheaSPK: But I'll just rewrite the test in order to test, if contacts etc. are not shown if not signed in

LinqLover:
@rheaSPK The ubuntu image is still in the repository. Apart from this, looks perfect now :-)

LinqLover (approved):
provided that you will remove the ubuntu log file before merging :)

frcroth:
@rheaSPK Thank you for your effort! I think the navigation can still be improved.

- What does the 4th button (the fa-history button) do / what is planned here?
- You can remove the dashboard navbar, as activities and proposals are not part of a scenario right now. This would remove the need for a second navbar

LinqLover:
@frcroth Interesting ideas, could you create a new issue for them?

frcroth:
OK https://github.com/hpi-swt2/connections-portal/issues/160

PR #148, opened 2021-01-22
Add handling of a log of type string

Alexander-Dubrawski:
There is an issue with the flash messages when logging out. The partial can only handle a dictionary with a heading and message. Somehow we are receiving a string instead of a dictionary while logging out. To fix this issue we simply check the type now. Not the best solution but for now it should work.

### DIFF
--- a/app/views/layouts/_flash_messages.html.erb
+++ b/app/views/layouts/_flash_messages.html.erb
@@ -1,12 +1,17 @@
 <% flash.each do |level, log| %>
     <div class='alert alert-<%= level %>'>
-      <% if !log[:heading].nil? %>
-        <h4><%= log[:heading] %></h4>
+      <% if log.is_a? String %>
+        <li><%= log %></li>
       <% end %>
-      <% log[:messages].each do |message| %>
-        <ul>
-          <li><%= message %></li>
-        </ul>
+      <% if !log.is_a? String %>
+        <% if !log[:heading].nil? %>
+          <h4><%= log[:heading] %></h4>
+        <% end %>
+        <% log[:messages].each do |message| %>
+          <ul>
+            <li><%= message %></li>
+          </ul>
+        <% end %>
       <% end %>
     </div>
 <% end %>


LinqLover:
Seems like a duplicate of 844735d in #120 as announced in the Discord ... 😫

LinqLover:
Well, let's decide on the better implementation and get it merged soon.

app/views/layouts/_flash_messages.html.erb
```diff
@@ -1,12 +1,17 @@
...
-      <% if !log[:heading].nil? %>
-        <h4><%= log[:heading] %></h4>
+      <% if log.is_a? String %>
+        <li><%= log %></li>
       <% end %>
-      <% log[:messages].each do |message| %>
-        <ul>
-          <li><%= message %></li>
-        </ul>
+      <% if !log.is_a? String %>
```
felixauringer, line 6:
```suggestion
      <% else %>
```

app/views/layouts/_flash_messages.html.erb
```diff
@@ -1,12 +1,17 @@
...
+        <% if !log[:heading].nil? %>
```
felixauringer, line 7:
```suggestion
        <% if log[:heading].present? %>
```

app/views/layouts/_flash_messages.html.erb
```diff
@@ -1,12 +1,17 @@
 <% flash.each do |level, log| %>
     <div class='alert alert-<%= level %>'>
-      <% if !log[:heading].nil? %>
-        <h4><%= log[:heading] %></h4>
+      <% if log.is_a? String %>
+        <li><%= log %></li>
       <% end %>
```
felixauringer, line 5:
```suggestion
```

felixauringer (changes requested):
Not sure if my if-else-end works with the suggestions but you get the idea.  
I still think that these flash messages are hacky. But this fix is more important than refactoring them.

felixauringer:
@LinqLover, you fixed this in [PR 120](https://github.com/hpi-swt2/connections-portal/pull/120), right? So I assume that this PR and branch can be deleted?

Alexander-Dubrawski:
Already fixed in #120

PR #149, opened 2021-01-22
Remove orange backgrounds from tables (we never have wanted this and …

LinqLover:
…also @chrisma did not like it)

Tiny PR, let's merge this one fast.

image 
image 
image


felixauringer (approved):

chrisma:
Don't worry about me not liking things... 😀
(However, consider the possibility that we might have added things to the starting code as an indicator of change 😀)

chrisma:
But, also, the orange is gone! 🥳

PR #117, opened 2021-01-12
Us97 add mobile support for contact page

rheaSPK:
Added a mobile support for the contact page

image

Closes #97.

PR #150, opened 2021-01-22
Add links to contacts again

LinqLover:
Originally implemented via #124, then accidentally removed via #117, here it is again 😅

Tiny PR, let's merge this one faaast.

### DIFF
--- a/app/views/contacts/index.html.erb
+++ b/app/views/contacts/index.html.erb
@@ -5,7 +5,7 @@
   <p><%= button_to 'View my requests', user_contact_requests_path(current_user), method: :get, :class => "btn btn-outline-primary"  %></p>
 
   <% @contacts.each do |contact| %>
-    <div class="card">
+    <a class="card" href=<% user_contacts_path(contact) %>>
       <div class="card-body">
         <div class="container">
           <div class="row">
@@ -30,7 +30,7 @@
           <% end %>
         </div>
       </div>
-    </div>
+    </a>
     <br>
   <%end %>
 <% end %>


felixauringer (approved):

frcroth:
This merge broke the CI for some reason

LinqLover:
Interesting, the same build has passed before pressing the merge button: https://github.com/hpi-swt2/connections-portal/actions/runs/504326073

LinqLover:
Even more interesting, 348ebc7383a29e2c4404b7760694a7a67864bb24 has fixed it again and after retrying https://github.com/hpi-swt2/connections-portal/actions/runs/504692213 it was green again, too. We appear to have at least one flaky test ... ❄️

felixauringer:
The reason for that is rather simple: Almost all of the user's attributes are generated randomly. The `user_search_spec.rb` (in `features`) searches for a part of the username and expects no or exactly one hit. If there is another name which also happens to contain the part that is searched for, there are more hits than the test expected. I already mentioned this in the [PR](https://github.com/hpi-swt2/connections-portal/pull/100#discussion_r559185070) but did not think that it will happen so quickly :sweat_smile:  I will write a fix for that.

felixauringer:
The reason for that is rather simple: Almost all of the user's attributes are generated randomly. The `user_search_spec.rb` (in `features`) searches for a part of the username and expects no or exactly one hit. If there is another name which also happens to contain the part that is searched for, there are more hits than the test expected. I already mentioned this in the [PR](https://github.com/hpi-swt2/connections-portal/pull/100#discussion_r559185070) but did not think that it will happen so quickly :sweat_smile:  I will write a fix for that.

PR #152, opened 2021-01-23
Add styling and additional logic to search for users

### LINKED ISSUE #98: [User Story] Search For Contacts In Mobile Version
# Description
As a user I want to search for contacts on my mobile, too.

# Dependencies
- [x] #97 

# Conditions of satisfaction
- [ ] the design follows the wireframe below
- [ ] there is a field to search for contacts
- [ ] only related contacts (= search results) are shown
- [ ] search results are alphabetically sorted

Mobile Contacts v2

### LINKED ISSUE #127: [Process Part] Search For Users
# Description
Als Nutzer möchte ich andere Benutzer über eine Suche finden können.

# Dependencies
- [ ] #126

# Conditions of satisfaction
- [ ] im Header gibt es ein Suchsymbol, auf das ich klicken kann und dann werden mir nur noch Benutzerkarten angezeigt, die zu der aktuellen Suche passen

Alexander-Dubrawski:
This PR closes #127 and closes #98 .

In the navbar is now an icon that is redirecting the user to the search page. 

The styling is now looking as follows: 

Desktop:
 
Screenshot 2021-01-29 at 22 31 56

Mobile: 

Screenshot 2021-01-29 at 22 08 54


app/controllers/users_controller.rb
```diff
@@ -44,12 +57,23 @@ def add_contact
...
   def search
-    @users = User.search(params[:search]).where.not(id: current_user.id)
-    @users_to_add = @users.reject do |user|
-      current_user.sent_contact_request?(user)
+    use_wildcards = true
+    @users = Set[]
+    @contacts = Set[]
+    if current_user && params[:search]
+      @user = User.find(current_user.id)
+      @users = search_record(params[:search], User, use_wildcards)
```
Paula-Kli: why are you not using the method from the Users model anymore? I don't think this is a good way to implement searching since you are using a method from the users helper but nevertheless have to tell which record you want to search in.
`User.search(params[:search]).where.not(id: current_user.id)` this line gave you all users which match the search parameter and which are not the current user

app/controllers/users_controller.rb
```diff
@@ -44,12 +57,23 @@ def add_contact
     redirect_to root_path
   end
 
+  # rubocop:disable Metrics/AbcSize
+  # rubocop:disable Metrics/MethodLength
   def search
-    @users = User.search(params[:search]).where.not(id: current_user.id)
-    @users_to_add = @users.reject do |user|
-      current_user.sent_contact_request?(user)
+    use_wildcards = true
```
Paula-Kli: if you are just turning that to `true` to give that to the method, why are you not putting `true` into the method call?

app/controllers/users_controller.rb
```diff
@@ -44,12 +57,23 @@ def add_contact
...
+    @users = Set[]
+    @contacts = Set[]
+    if current_user && params[:search]
+      @user = User.find(current_user.id)
```
Paula-Kli: I think that variable could be named `@current_user` to be more precise

app/views/layouts/_flash_messages.html.erb
```diff
@@ -1,7 +1,10 @@
 <% flash.each do |level, log| %>
     <div class='alert alert-<%= level %>'>
-      <% if log.respond_to?(:each) %>
-        <% if log[:heading].nil? %>
+      <% if log.is_a? String %>
```
Paula-Kli: what does this change mean? why is this changed in the "change search" pr?

app/views/users/search.html.erb
```diff
@@ -1,42 +1,51 @@
...
+  <% end %>
 
-
-    <div class="row">
-        <div class="col">
-          <%= form_tag(search_users_path, method: "get") do %>
-            <%= label_tag(:search, I18n.t('user.search.for') ) %>
-            <%= text_field_tag :search, params[:search] %>
-            <%= submit_tag I18n.t('user.search.index') %>
+  <% @contacts.each do |user| %>
```
Paula-Kli: is there a reason, why the logic is changed? should I first see my contacts and later other people?

I would not call one contact just user but `contact` to know why they are handled differently

when #158  ist merged you can just call `<%= render partial: 'users/card', locals: { user: contact, mode: :overview } %>` here to get the expected card with the social accounts

app/views/users/search.html.erb
```diff
@@ -1,42 +1,51 @@
...
-            <td><%= user.username%></td>
-            <td><%= user.firstname%></td>
-            <td><%= user.lastname%></td>
-            <td><%= user.email %></td>
-            <% if @users_to_add.include? user %>
-              <td><%=  button_to '+', user_contact_requests_path(user), method: :post, id: user.id  %></td>
-            <% end %>
-          </tr>
+  <% @users.each do |user| %>
+    <div class=contact>
```
Paula-Kli: when #158  ist merged you can simply use `<%= render partial: 'users/card', locals: { user: user, mode: :add } %>` to get the expected card with a `+` button and a tooltip

config/routes.rb
```diff
@@ -5,6 +5,7 @@
   devise_for :users
   # For details on the DSL available within this file, see https://guides.rubyonrails.org/routing.html
   root to: 'home#dashboard'
+  get 'search', to: 'users#search'
```
Paula-Kli: why is this line added when the same is done in line 14?

app/controllers/users_controller.rb
```diff
@@ -44,12 +57,23 @@ def add_contact
...
+      @users = search_record(params[:search], User, use_wildcards)
+      @contacts = search_record(params[:search], @user.contacts, use_wildcards)
```
Paula-Kli: please use `current_user.sent_contact_request?` here because we not only want to send no second request to a user that is already our contact but also not to someone with a pending request

Paula-Kli:
should the empty search page maybe display all available users?

app/controllers/users_controller.rb
felixauringer: Because it is currently not necessary to switch wildcards off, I would completely omit this variable and always use wildcards.

app/controllers/users_controller.rb
```diff
@@ -44,12 +57,23 @@ def add_contact
     redirect_to root_path
   end
 
+  # rubocop:disable Metrics/AbcSize
```
felixauringer: I would prefer changing code not the config of rubocop. Why should we use it at all if we are turning it off when it detects opportunities for improvements?

app/assets/stylesheets/users.scss
```diff
@@ -1,3 +1,86 @@
...
+     font-size: small;
+}
+.add_button {
+    flex: 1;
+    height: 100%;
+    width: 100%;
+    border-radius: 99999px;
+    background: #343B3F;
+    font-size: small;
+}
```
felixauringer: missing newline

app/controllers/users_controller.rb
```diff
@@ -1,9 +1,18 @@
+require 'set'
+
 class UsersController < ApplicationController
   include SocialAccountsHelper
-  before_action :authorize
-  helper_method :generate_link, :supported_social_networks
-
-  def show; end
+  include UsersHelper
+  before_action :authorize, except: %i[show index search]
```
felixauringer: Why do you change the authorization settings for this controller? I couldn't find it specified in the linked ticket and I think @LinqLover did really good work when introducing proper authentication.

app/controllers/users_controller.rb
felixauringer: I would strongly support this. The search logic built by Scrumbledore's Army is reviewed and well-tested. I do not see any reason to rebuild this logic. And I would also prefer the approach of the existing solution over your new approach.

app/helpers/users_helper.rb
```diff
@@ -1,2 +1,25 @@
+require 'set'
+
 module UsersHelper
+  # rubocop:disable Metrics/MethodLength
```
felixauringer: You can't just disable Rubocop wheresoever it complains about something :sweat:

app/views/layouts/_navbar.html.erb
```diff
@@ -3,6 +3,7 @@
   <div class="navbar-brand">Connection Portal</div>
   <form class="form-inline">
     <i class="fa fa-bell" style="color: white; padding: 10px"></i>
-    <a href="<%= search_users_path %>"><i class="fa fa-search" style="color: white; padding: 10px"></i> </a>
+    <%= link_to (fa_icon 'search'), "/search", style:  "color: white; padding: 10px"%>
```
felixauringer: please use path helpers

app/views/users/search.html.erb
felixauringer: That is now merged :+1:

app/views/users/search.html.erb
```diff
@@ -1,42 +1,51 @@
...
+      <label class="contact-info">
+        <div class="profile-image my-2 my-sm-0 ">
```
felixauringer: This is almost the same code as for contacts above

felixauringer (changes requested):
Remove the `.DS_Store` file please. You could also add it to your global gitignore on your computer (not in the project's gitignore) to prevent committing this file in all git projects you are working on.

You should merge `dev` into this branch because it already contains a lot of changes that you tried to implement here. I would always wait for PRs to be merged instead of implementing them yourself. That duplicates work and increases the size of PRs so reviewing gets a lot harder.

So I think all the things mentioned in `Additional implementation` in the PR description should not be in this branch. For example the logout button: If you notice that it is missing, you could open a ticket, fix it with maybe 5 lines of code, write one test for it and the PR could be merged in one day. But if you include something like this in a PR as big as this one, it takes forever until users can log out again.

app/controllers/users_controller.rb
Alexander-Dubrawski: Hm... You are right. I don't know what happened there 🤔 . I will use the search logic from before.

app/controllers/users_controller.rb
Alexander-Dubrawski: I replaced the search logic with the one from Scrumbledore's Army. So we are not using this variable anymore.

app/controllers/users_controller.rb
Alexander-Dubrawski: Same as above.

app/views/layouts/_flash_messages.html.erb
Alexander-Dubrawski: It was just a quick fix to be able to continue implementing. But after the merge with the master, I replaced the code with the new code from there.

app/views/users/search.html.erb
Alexander-Dubrawski: > is there a reason, why the logic is changed? should I first see my contacts and later other people?

Yes, that is the desired behavior.

Alexander-Dubrawski: I would like to use the new design, especially since it is a pretty nice generic solution. But after some testing (Iphone 6/7/8), I am a little bit conflicted. The links don't seem to be placed correctly:  

Screenshot 2021-02-03 at 20 59 19

I just changed the code to:

```rails
      <% @users.each do |user| %>
             <%= render partial: 'users/card', locals: { user: user, mode: :overview } %>
     <% end %>
```

Am I doing something wrong? 😞

app/views/users/search.html.erb
Alexander-Dubrawski: Please have a look at the conversation above.

config/routes.rb
Alexander-Dubrawski: We need this route if we try to reload the search page. Not sure why it is not working with the one defined in line 14 though, to be honest.

app/controllers/users_controller.rb
Alexander-Dubrawski: Removed this search logic.

app/controllers/users_controller.rb
Alexander-Dubrawski: In my opinion, sometimes methods will be more complex than 10 lines of code. But I am programming quite a lot of c++ lately so maybe I just have a different subjective view on this matter. Since I replaced this method with the one of  Scrumbledore's Army I don't need to disable rubocop here.

app/controllers/users_controller.rb
Alexander-Dubrawski: Not sure why I did it. I removed it.

app/views/users/search.html.erb
Alexander-Dubrawski: I know 🙈 . Just wanted to wait for the implementation of  #158 . Otherwise, I would have implemented some genric partials.

app/views/users/search.html.erb
Alexander-Dubrawski: If you can't find a quick fix I could try to adjust it with some CSS.

Paula-Kli: I dont think you are doing somethin wrong and for me with firefox and iPhone 6/7/8 it looks like this:
grafik

Paula-Kli: I changed the code, to the desired partial but I am on your branch

config/routes.rb
Paula-Kli: did you try using the path helper `search_users_path` as mentioned below it is not just `/search` but `users/search` and due to that the correct path is `search_users_path`

app/views/layouts/_flash_messages.html.erb
```diff
@@ -13,4 +13,4 @@
         <h4><%= log %></h4>
       <% end %>
     </div>
-<% end %>
+<% end %>
```
rheaSPK: Please add new line

app/views/layouts/_footer.html.erb
```diff
@@ -1,29 +1,58 @@
+<style>
```
rheaSPK: please put this a seperate css file like `mobile.scss`

rheaSPK (commented):
Please add the new card style, it's already merged and please fix CI-Errors, somethings not right with some of your code...

app/views/layouts/_footer.html.erb
```diff
@@ -1,29 +1,58 @@
...
-        </a></li>
+        <label class="nav-bar-entry">
+          <a class="nav-link <%= active_link_to_class(edit_user_path(current_user), active: :exclusive)%>" href="<%= edit_user_path(current_user) %>">
+            <i class="fa fa-cog"></i>
+          </a>
+        </label>
+      </li>
     </ul>
   </div>
 <% end %>
```
felixauringer: missing newline

app/views/layouts/_navbar.html.erb
```diff
@@ -3,7 +3,8 @@
   <div class="navbar-brand"><%= link_to I18n.t('index.portal_name'), root_path, class: 'portal-link' %></div>
   <form class="form-inline">
     <i class="fa fa-bell" style="color: white; padding: 10px"></i>
-    <a href="<%= search_users_path %>"><i class="fa fa-search" style="color: white; padding: 10px"></i> </a>
+    <%= link_to (fa_icon 'search'), "/search", style:  "color: white; padding: 10px"%>
+    <%= link_to (fa_icon 'sign-out'), destroy_user_session_path, method: :delete, class: 'nav-link' %>
   </form>
 <%= render 'layouts/navbar_desktop_part' %>
 </nav>
```
felixauringer: missing newline

app/views/users/search.html.erb
felixauringer: #158 has been merged now

spec/features/user_search_spec.rb
```diff
@@ -7,9 +7,16 @@
 
   describe 'without filter' do
     let!(:users) { FactoryBot.create_list(:user, 3) }
+    let(:second_user) { FactoryBot.create :user }
 
     before { visit search_users_path }
 
+    it 'is possible to filter by username' do
```
felixauringer: Why did you add this test? Exactly this behaviour is tested in the line 55 (before your changes) / 62 (after your changes).
If you want to check that the first name is visible, then I would add this expectation to all tests below

spec/views/users/search.html.erb_spec.rb
```diff
@@ -2,15 +2,23 @@
...
-    assign(:users, users)
+    assign(:users, contacts)
     assign(:users_to_add, users)
-    sign_in users.first
+    sign_in user1
     render
   end
 
-  it 'renders a list of users before searching' do
+  it 'renders a list of contacts' do
```
felixauringer: What's your thought behind the changes in this test? You have renamed the variable in the test and added the existing test again with the renamed variable but that shouldn't change the behaviour of the test. I would therefore restore this file to its current state on dev

app/assets/stylesheets/users.scss
```diff
@@ -1,3 +1,86 @@
 // Place all the styles related to the Users controller here.
 // They will automatically be included in application.css.
 // You can use Sass (SCSS) here: https://sass-lang.com/
+
```
felixauringer, line 4: Is this file still necessary with #158 merged?

app/controllers/users_controller.rb
```diff
@@ -1,9 +1,14 @@
+require 'set'
```
felixauringer: why?

app/controllers/users_controller.rb
```diff
@@ -1,9 +1,14 @@
...
+
 class UsersController < ApplicationController
   include SocialAccountsHelper
-  before_action :authorize
+  include UsersHelper
+  before_action except: %i[show index search]
```
felixauringer: What does this do? I would expect the symbol of a method name here that would be executed before the specified actions

app/controllers/users_controller.rb
```diff
@@ -1,9 +1,14 @@
+require 'set'
+
 class UsersController < ApplicationController
   include SocialAccountsHelper
-  before_action :authorize
+  include UsersHelper
```
felixauringer: ?
```suggestion
```

app/controllers/users_controller.rb
```diff
@@ -1,9 +1,14 @@
...
+  before_action except: %i[show index search]
   helper_method :generate_link, :supported_social_networks
 
-  def show; end
+  def show
```
felixauringer: This change should not be necessary if you would leave `before_action :authorize` this would not be necessary

app/controllers/users_controller.rb
```diff
@@ -31,6 +36,10 @@ def update_status
   end
 
   def index
+    unless current_user
+      redirect_to(new_user_session_path)
```
felixauringer: This change should not be necessary if you would leave before_action :authorize this would not be necessary

app/views/users/search.html.erb
Alexander-Dubrawski: @Paula-Kli there are a few problems with the cards that catch my eye:

* Why are we using tables to style the cards? Tables were used in the early days of CSS. Now usually flex or grids are used for that. With flex, you are able to place the content much nicer and evenly. For example, you can distribute the handles evenly. I think the problem that we saw is caused by using tables.
* Why are we using cards from Bootstrap. Bootstrap cards usually have a different layout. If we want to simply display a shadow and have round corners we can  use basic CSS

I could quickly adjust the code of your partials so that the problems are solved if it is ok for you 🙃 . We already have all the CSS.

app/assets/stylesheets/users.scss
Alexander-Dubrawski: Yes, it is. We are using some styles for the search form.

config/locales/en.yml
```diff
@@ -71,6 +71,7 @@ en:
     search:
       index: 'Search'
       for: 'Search for:'
+      place_holder: 'Username, email, ...'
```
frcroth: Use placeholder instead of place_holder

app/views/layouts/_footer.html.erb
```diff
@@ -1,29 +1,60 @@
...
+
 <% if user_signed_in? %>
   <div id="navigation">
     <ul class="nav nav-pills nav-justified">
       <li class="nav-item">
-        <a class="nav-link <%= active_link_to_class(root_path, active: :exclusive)%>" href="<%= root_url %>">
-          <i class="fa fa-home"></i>
-        </a></li>
+         <label class="nav-bar-entry">
+          <!-- new path to this when dashboard is rewritten -->
```
frcroth: This comment does not seem valid

app/views/layouts/_footer.html.erb
```diff
@@ -1,29 +1,60 @@
...
+            <a class="nav-link footer-nav-link <%= active_link_to_class(root_path, active: :exclusive)%>" href="<%= root_url %>">
+            <i class="fa fa-home"></i>
+          </a>
+        </label>
+      </li>
       <li class="nav-item">
-        <a class="nav-link <%= active_link_to_class(chat_path, active: :exclusive)%>" href="<%= chat_path %>">
-          <i class="fa fa-comments"></i>
-        </a></li>
+        <!-- please update the path when reworked the chat -->
```
frcroth: Remove

app/views/layouts/_footer.html.erb
```diff
@@ -1,29 +1,60 @@
...
+          <label class="nav-bar-entry">
+          <a class="nav-link footer-nav-link <%= active_link_to_class(root_path, active: :exclusive)%>" href="<%= chat_path %>">
```
frcroth: Replace root_path with chat_path

app/views/layouts/_footer.html.erb
```diff
@@ -1,29 +1,60 @@
...
-        <a class="nav-link <%= active_link_to_class(edit_user_path(current_user), active: :exclusive)%>"
-           href="<%= edit_user_path(current_user) %>">
-          <i class="fa fa-cog"></i>
-        </a></li>
+        <label class="nav-bar-entry">
+          <a class="nav-link footer-nav-link <%= active_link_to_class(edit_user_path(current_user), active: :exclusive)%>" 
+            href="<%= edit_user_path(current_user) %>">
+            <i class="fa fa-cog"></i>
+          </a>
+        </label>
```
frcroth: What is the reason for the changes in this file? It does not seem to be related to the topic of the PR. You could remove this here/ move it to another PR to reduce complexity

frcroth (changes requested):

app/views/layouts/_footer.html.erb
Alexander-Dubrawski: Maybe the following screenshot describes it the best: 

Screenshot 2021-02-05 at 17 09 16

The link of the setting page is covering the whole page.

Alexander-Dubrawski: To solve this problem I simply wrap the link inside a label.

frcroth: Sorry I don't have access to a PC right now. Can you reproduce this @felixauringer?

config/routes.rb
Alexander-Dubrawski: Now it works, thanks 😄 .

spec/views/users/search.html.erb_spec.rb
Alexander-Dubrawski: We have *users* and *users_to_add* and we want to check that both groups are being rendered. That's why we have two test cases.

spec/features/user_search_spec.rb
Alexander-Dubrawski: I removed the test.

app/controllers/users_controller.rb
Alexander-Dubrawski: I added `before_action :authorize` again.

app/controllers/users_controller.rb
Alexander-Dubrawski: I changed it back

app/controllers/users_controller.rb
Alexander-Dubrawski: changed it back

app/views/layouts/_footer.html.erb
Alexander-Dubrawski: @frcroth is it okay if I leave it in this PR? I know it's not the best solution, but at the moment I prefer the fastest one.

spec/features/user_search_spec.rb
```diff
@@ -7,6 +7,7 @@
 
   describe 'without filter' do
     let!(:users) { FactoryBot.create_list(:user, 3) }
+    let(:second_user) { FactoryBot.create :user }
```
felixauringer: Is this user still needed? It is the only change in this file, so I assume that it can be omitted?
```suggestion
```

spec/views/users/search.html.erb_spec.rb
felixauringer: That's a good idea, but that is not what the test does :sweat_smile:

felixauringer: `assign(:users, contacts)` is like executing `@users = contacts` in the controller before rendering the view (    assign(:users, contacts))

spec/views/users/search.html.erb_spec.rb
```diff
@@ -2,15 +2,23 @@
 
 RSpec.describe 'users/search', type: :view do
   let(:users) { FactoryBot.create_list(:user, 3) }
+  let(:contacts) { FactoryBot.create_list(:user, 3) }
+  let(:user1) { FactoryBot.create :user }
```
felixauringer:
```suggestion
  let(:user) { FactoryBot.create :user }
```

spec/views/users/search.html.erb_spec.rb
```diff
@@ -2,15 +2,23 @@
...
 
   before do
-    assign(:users, users)
+    assign(:users, contacts)
```
felixauringer:
```suggestion
    assign(:users, contacts + users)
```

spec/views/users/search.html.erb_spec.rb
```diff
@@ -2,15 +2,23 @@
...
     assign(:users_to_add, users)
-    sign_in users.first
+    sign_in user1
```
felixauringer:
```suggestion
    sign_in user
```

felixauringer (commented):
I added some suggestions that assign the instance variables in the view as the users controller would do

app/views/layouts/_footer.html.erb
felixauringer: image

I'm on your branch @Alexander-Dubrawski but I resetted `app/views/layouts/_footer.html.erb` to its state on the `dev` branch and everything seems to be fine without your changes in this file. I'm not quite sure what the problem is :sweat_smile:

Alexander-Dubrawski: That's confusing. I have the same problem on dev.  But I will double-check.

Alexander-Dubrawski: Well..., somehow it got magically fixed. I can't reproduce it anymore either. Maybe one of the last merges fixed it. Its getting hard to keep track of.

spec/views/users/search.html.erb_spec.rb
Alexander-Dubrawski: I just changed the code back to the dev status.

app/views/users/search.html.erb
Alexander-Dubrawski: will be solved with #189 .

felixauringer:
The code looks great now! The PR has also a size that is very good reviewable. Just one question:
```ruby
  def search
    @users = User.search(params[:search]).where.not(id: current_user.id)
    @users_to_add = @users.reject do |user|
      current_user.sent_contact_request?(user)
    end
  end
```
It looks like `users_to_add` is a subset of `users`. Did you try out whether users that aren't contacts of (or requested by) the current user are displayed twice?
If yes, I would suggest something like the following
```ruby
  def search
    other_users = User.search(params[:search]).where.not(id: current_user.id)
    @users_to_add = other_users.reject do |user|
      current_user.sent_contact_request?(user)
    end
    @users = other_users.select do |user|
      current_user.sent_contact_request?(user)
    end
  end
```

Alexander-Dubrawski:
> The code looks great now! The PR has also a size that is very good reviewable. 

Sorry for this chaotic PR of mine 🙈 . I really don't know why I implemented stuff that was already existing. But thanks for all your feedback and patience. 
  

> It looks like `users_to_add` is a subset of `users`. Did you try out whether users that aren't contacts of (or requested by) the current user are displayed twice?

You are absolutely right. Good eye. I adjusted the code.

felixauringer (approved):
Nice to finally finish this :tada: I'm sorry for having so much comments but thank you for implementing all of it :)

frcroth:
image
This is what /users/search looks for me on the deployed version

frcroth:
@Alexander-Dubrawski

Alexander-Dubrawski:
> This is what /users/search looks for me on the deployed version

I can not reproduce the error. @frcroth what exactly did you do? (For example: login, click the search icon, ...)

felixauringer:
> image
> This is what /users/search looks for me on the deployed version

Same for me, I will try to investigate the reason

Alexander-Dubrawski:
> Same for me, I will try to investigate the reason

@felixauringer thanks that would be really nice. If you can not find the error, please ping me and I can also have a look this evening.

felixauringer:
Works locally for me, so I will have a look at the heroku logs

felixauringer:
The relevant part is

```text
ActionView::Template::Error (Can't resolve image into URL: to_model delegated to attachment, but attachment is nil):

2021-02-06T15:00:46.081546+00:00 app[web.1]:      3:     <div class="container">

2021-02-06T15:00:46.081546+00:00 app[web.1]:      4:       <div class="row">

2021-02-06T15:00:46.081547+00:00 app[web.1]:      5:         <div class="col-xs">

2021-02-06T15:00:46.081548+00:00 app[web.1]:      6:           <%= image_tag user.avatar, height: 65, width: 65, class: "rounded-circle" %>

2021-02-06T15:00:46.081548+00:00 app[web.1]:      7:         </div>

2021-02-06T15:00:46.081554+00:00 app[web.1]:      8:         <a class="col-8" href=<%= user_path(user) %>>

2021-02-06T15:00:46.081555+00:00 app[web.1]:      9:           <h5 class="card-title"><%= user.display_name %></h5>

2021-02-06T15:00:46.081556+00:00 app[web.1]:   

2021-02-06T15:00:46.081556+00:00 app[web.1]:  app/views/users/_card.html.erb:6

2021-02-06T15:00:46.081556+00:00 app[web.1]:  app/views/users/search.html.erb:14

2021-02-06T15:00:46.081557+00:00 app[web.1]:  app/views/users/search.html.erb:13:in `each'

2021-02-06T15:00:46.081557+00:00 app[web.1]:  app/views/users/search.html.erb:13
```

felixauringer:
So again problems with the avatars. This PR did not touch `_card.html.erb` so this problem was likely there before merging this one

felixauringer:
That's executed on production data:

```ruby
User.all.select { |user| user.avatar.nil? }.length
D, [2021-02-06T15:08:13.259421 #4] DEBUG -- :   User Load (1.8ms)  SELECT "users".* FROM "users"
=> 0
```
So the migration worked as expected

felixauringer:
I added an issue for this: #191

PR #153, opened 2021-01-25
Start jisti calls

### LINKED ISSUE #42: Option to start jitsi meeting
## User Story

**As a** user -  
**I want to** be able to start a jitsi meeting -  
**In order to** speak with other people on the platform.

## Conditions of satisfaction

- [x] a new random jitsi meeting id can be automatically generated
- [x] the meeting id should be assigned to two users:
  - [x] user who initiated creation
  - [x] user the meeting should be initialized with
- [x] when clicking a certain button associated with another user, setup meeting id with that user

## Implementation Note

Hard-code the url from a server where meetings can be hosted on.

frcroth:
Closes #42


spec/features/jitsi_calls_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    sign_in user
+    visit root_path
+  end
+
+  it 'creates call when clicking on button' do
+    within("#init-call-#{participant.id}") do
+      click_button('button')
+      new_call = user.jitsi_calls.first
+      expect(new_call.call_participants.map(&:user_id)).to include(user.id)
+      expect(new_call.call_participants.map(&:user_id)).to include(participant.id)
```
felixauringer: I would also write a test to check whether the roles are correct.

spec/models/user_spec.rb
```diff
@@ -138,4 +140,25 @@
       expect(user.display_name).to eq('hasso01')
     end
   end
+
+  context 'with jitsi calls' do
+    let(:jitsi_calls) { FactoryBot.create_list :jitsi_call, 2 }
+
+    before do
+      user.save
```
felixauringer, line 148: We wouldn't need this if we use build in the next line, would we?

spec/models/user_spec.rb
```diff
@@ -138,4 +140,25 @@
...
+      jitsi_calls.each { |j| user.call_participants.create(jitsi_call: j, state: 'accepted', role: 'participant') }
```
felixauringer: Because the tests do not depend on a specific state or role, we could use the first one from their valid lists so we do not have to change the tests if the valid lists change.

spec/models/jitsi_call_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  it 'is invalid without a room name' do
+    jitsi_call.room_name = ''
+    expect(jitsi_call).not_to be_valid
+  end
+
+  context 'with participants' do
+    let(:participants) { FactoryBot.create_list :user, 2 }
+
+    before do
+      participants.each { |p| jitsi_call.call_participants.create(user: p, state: 'accepted', role: 'participant') }
```
felixauringer: Because the tests do not depend on a specific state or role, we could use the first one from their valid lists so we do not have to change the tests if the valid lists change.

spec/requests/jitsi_call_request_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rails_helper'
+
+RSpec.describe 'Jitsi Call', type: :request do
+  describe 'POST' do
+    context 'when logged in' do
+      let(:participant) { FactoryBot.create(:user) }
+
+      before do
+        sign_in FactoryBot.create(:user)
+      end
```
felixauringer:
```suggestion
      before { sign_in FactoryBot.create(:user) }
```

felixauringer (commented):
Just some first thoughts.

app/controllers/jitsi_calls_controller.rb
```diff
@@ -0,0 +1,22 @@
+class JitsiCallsController < ApplicationController
+  before_action :authenticate_user!
+
+  def create
+    room_name = SecureRandom.uuid
+    @jitsi_call = JitsiCall.new({ room_name: room_name })
```
felixauringer:
```suggestion
    @jitsi_call = JitsiCall.new(room_name: SecureRandom.uuid)
```

app/controllers/jitsi_calls_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    if @jitsi_call.save && User.exists?(call_params[:user_id])
+      assign_participants
+    else redirect_to root_path, alert: I18n.t('denial.resource_creation', resource: JitsiCall) end
```
felixauringer: I'm very confused that RuboCop does not forbid `else ... end` in one line :sweat_smile:

app/controllers/jitsi_calls_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def assign_participants
+    @jitsi_call.call_participants.create({ state: 'accepted', role: 'initiator', user: current_user })
```
felixauringer: if you give one hash and no other parameters to a function, you can omit the curly braces.

app/views/home/_users.html.erb
```diff
@@ -15,7 +15,11 @@
         <%= user.display_name %>
       </div>
       <div class="user-meeting d-flex">
-        <span><i class="fa fa-phone fa-flip-horizontal"></i></span>
+        <%= form_with model: JitsiCall.new, id: "init-call-#{user.id}" do |f| %>
+          <%= f.hidden_field :user_id, value: user.id %>
+          <%= button_tag(type: 'submit', class: "fa fa-phone fa-flip-horizontal phone-button") do %>
```
felixauringer, line 20: Have you tried `f.submit` here? Maybe one can omit the `type` then

spec/requests/jitsi_call_request_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+      it 'does not create a new call without participant user id' do
+        post jitsi_calls_path, params: { jitsi_call: { user_id: nil } }
+        expect(response).to redirect_to(root_path)
+      end
+    end
+
+    context 'without logging in' do
+      it 'redirects to login page' do
+        post jitsi_calls_path, params: { jitsi_call: { user_id: 0 } }
```
felixauringer: I think you should use `participant.id` here too. Then it would be clear that the redirect is because of authentication and not because the not found logic maybe performs some weird redirect.

felixauringer (changes requested):

app/views/home/_users.html.erb
frcroth: Yes this results in an awful button

frcroth:
@felixauringer I have no idea about the model stuff, you can change these things

sirkrypt0 (approved):

spec/requests/jitsi_call_request_spec.rb
frcroth: Fixed this in 49cad28

app/controllers/jitsi_calls_controller.rb
frcroth: I removed the `end`

spec/models/user_spec.rb
frcroth: No idea

spec/models/user_spec.rb
frcroth: Fixed in 2b5f129

spec/models/jitsi_call_spec.rb
frcroth: Fixed in 2b5f129. I also added a new do end thingy around it because the line got to long. Maybe you can do it better.

spec/models/user_spec.rb
felixauringer: It does not work because we check the number of call participants in the database in `destroys all participants when destroyed`

spec/features/jitsi_calls_spec.rb
felixauringer: Konrad implemented this.

felixauringer (approved):

PR #154, opened 2021-01-25
Add migration that adds global chat as seeds are not used in deployment

sirkrypt0:


felixauringer (approved):

mpass99:
Don't we have the problem here again that the db:setup for the Github Actions doesn't execute the migrations (correctly)?!

PR #155, opened 2021-01-25
Configure production to use async websocket adapter and allow origins

sirkrypt0:


felixauringer (approved):
Not quite sure what effects a change of the adapter could have but we'll see :D

sirkrypt0:
> Not quite sure what effects a change of the adapter could have but we'll see :D

Should be fine as it hasn't been used before

Issue #47, opened 2020-12-02
Settings Page


florian-papsdorf:
As a user I want to have one single page where I can edit all settings of the app.

LinqLover:
Do we already have any settings or is this one still blocked?

Edit: Okay, I just found #48. Maybe it would be a good idea to reference these issues between each other so we don't loose the overview right now. 😅

florian-papsdorf:
![Account Settings](https://user-images.githubusercontent.com/49521681/106022056-bcf61600-60c5-11eb-84de-5fef3350da18.png)

LinqLover:
@florian-papsdorf A few questions:

- Do I understand correctly that the views for user settings & profile edit should be merged?
- There is no wireframe for changing the profile status or editing your name/email address. How should this look like? Redirect to another page? Also, does the status really belong to the settings? There is an option for this right on the dashboard, isn't it?
- What should the eye button do?
- Don't we need two input fields for each new handle, one for the platform and another one for the handle?

LinqLover:
Answers to the questions:

> * Do I understand correctly that the views for user settings & profile edit should be merged?

Yes

> * There is no wireframe for changing the profile status or editing your name/email address. How should this look like? Redirect to another page? Also, does the status really belong to the settings? There is an option for this right on the dashboard, isn't it?

- Show a modal for name & email address when the edit button is clicked
- Show a modal for choosing the user picture when the picture is clicked
- We won't show or edit the status in the settings

> * What should the eye button do?

- Hide or show the selected handle (new functionality!)

> * Don't we need two input fields for each new handle, one for the platform and another one for the handle?

- Click on the question mark icon to get a modal for choosing the platform. When the form is submitted earlier, I think we should show the same modal, too.

LinqLover:
Task structure:

- [ ] Navbar: Remove link to Account Settings
- [x] New view/route (`edit_profile`) to change name, email address, and place from same controller (extract from user edit + account settings) (#157)
- [x] Remove email address from account settings view (#157)
- [ ] Redesign user edit page:
  - [ ] show (dummy) picture
  - [ ] remove status editor
  - [ ] new CSS for page + social handles
  - [ ] add links to pages for changing password, deleting account, and logging out
  - [ ] add modals for editing name, email & place; editing social handle name; and choosing social handle platform

PR #157, opened 2021-01-29
Separate edit profile page

LinqLover:
Part of #47, see https://github.com/hpi-swt2/connections-portal/issues/47#issuecomment-769821068.

Tasks:
- [x] New view/route (`edit_profile`) to change name, email address, and place from same controller (extract from user edit + account settings)
- [x] Link to it from edit user page
- [x] Remove email address from account settings view

In particular note the monster commit cf9060f and the relevant commit message ...


Paula-Kli:
what is the difference between /edit and /edit_profile? is there a particular reason why they are not the same page?

LinqLover:
> what is the difference between /edit and /edit_profile? is there a particular reason why they are not the same page?

Yes, this is a part of #47, where we want to have a settings/profile page with an edit link which then leads you to the edit_profile page. Design will be worked out in another PR, but in the end the edit page will show up a modal dialog with the edit_profile form when requested.

app/views/users/edit.html.erb
```diff
@@ -1,40 +1,7 @@
 <h1><%= I18n.t('user.edit.edit_profile_heading') %></h1>
```
felixauringer, line 1: This would not be the `Edit profile page` then, would it?

app/views/users/mailer/confirmation_instructions.html.erb
```diff
@@ -1,5 +1,6 @@
-<p>Welcome <%= @email %>!</p>
+<p><%= t('.greeting', :recipient => @resource.email, :default => "Welcome #{@resource.email}!") %></p>
```
felixauringer, line 1: why is this default here in the code? I would have expected this string in the English locales and the English locales set as the fallback locale. This should have the same effect but the code would be much cleaner

app/views/users/mailer/confirmation_instructions.html.erb
```diff
@@ -1,5 +1,6 @@
...
 
-<p>You can confirm your account email through the link below:</p>
+<p><%= t('.instruction', :default => "You can confirm your account email through the link below:") %></p>
 
-<p><%= link_to 'Confirm my account', confirmation_url(@resource, confirmation_token: @token) %></p>
+<p><%= link_to t('.action', :default => "Confirm my account"),
+         confirmation_url(@resource, :confirmation_token => @resource.confirmation_token) %></p>
```
felixauringer: missing newline

We really need a linter for `.html.erb` :smile:

app/views/users/mailer/reset_password_instructions.html.erb
```diff
@@ -1,8 +1,8 @@
...
-<p>Someone has requested a link to change your password. You can do this through the link below.</p>
+<p><%= t('.instruction', :default => "Someone has requested a link to change your password, and you can do this through the link below.") %></p>
 
-<p><%= link_to 'Change my password', edit_password_url(@resource, reset_password_token: @token) %></p>
+<p><%= link_to t('.action', :default => "Change my password"), edit_password_url(@resource, :reset_password_token => @resource.reset_password_token) %></p>
 
-<p>If you didn't request this, please ignore this email.</p>
-<p>Your password won't change until you access the link above and create a new one.</p>
+<p><%= t('.instruction_2', :default => "If you didn't request this, please ignore this email.") %></p>
+<p><%= t('.instruction_3', :default => "Your password won't change until you access the link above and create a new one.") %></p>
```
felixauringer: missing newline

app/views/users/mailer/unlock_instructions.html.erb
```diff
@@ -1,7 +1,7 @@
...
+<p><%= t('.greeting', :recipient => @resource.email, :default => "Hello #{@resource.email}!") %></p>
 
-<p>Your account has been locked due to an excessive number of unsuccessful sign in attempts.</p>
+<p><%= t('.message', :default => "Your account has been locked due to an excessive amount of unsuccessful sign in attempts.") %></p>
 
-<p>Click the link below to unlock your account:</p>
+<p><%= t('.instruction', :default => "Click the link below to unlock your account:") %></p>
 
-<p><%= link_to 'Unlock my account', unlock_url(@resource, unlock_token: @token) %></p>
+<p><%= link_to t('.action', :default => "Unlock my account"), unlock_url(@resource, :unlock_token => @resource.unlock_token) %></p>
```
felixauringer, line 7: I would prefer the new style of ruby hashes (`{ default: 'test' }` instead of `{ :default => 'test' }`) in all the html files

app/views/users/shared/_links.erb
```diff
@@ -0,0 +1,25 @@
...
+
+<%- if devise_mapping.lockable? && resource_class.unlock_strategy_enabled?(:email) && controller_name != 'unlocks' %>
+<%= link_to t('.didn_t_receive_unlock_instructions', :default => "Didn't receive unlock instructions?"), new_unlock_path(resource_name) %><br />
+<% end -%>
+
+<%- if devise_mapping.omniauthable? %>
+  <%- resource_class.omniauth_providers.each do |provider| %>
+  <%= link_to t('.sign_in_with_provider', :provider => provider.to_s.titleize, :default => "Sign in with #{provider.to_s.titleize}"), omniauth_authorize_path(resource_name, provider) %><br />
+  <% end -%>
+<% end -%>
```
felixauringer: missing newline

config/locales/devise.en.yml
```diff
@@ -64,3 +99,12 @@ en:
...
         other: "%{count} errors prohibited this %{resource} from being saved:"
+  activerecord:
+    models:
+      user: "User"
+    attributes:
+      user:
+        current_sign_in_at: "Signed in at"
+        email: "Email"
+        password: "Password"
+        last_sign_in_at: "Last signed in at"
```
felixauringer: missing newline

felixauringer (changes requested):
Nice! I always wondered how to customize devise views :sweat_smile: 

I find this PR really hard to review. Would it also be possible to customize only some views of devise?

rheaSPK (approved):
Looks fine to me (except some minor things @felixauringer mentioned), thanks for the reconfiguration, must have been annoying ^^

LinqLover:
@felixauringer Thanks for your review! All your suggestions but the first one refer to auto-generated code (I did not write it by hand ...). Do you think it is a good idea to change these auto-generations? Otherwise, I'd just leave it as-is.

app/views/users/edit.html.erb
LinqLover: Resolved via e2b32193a7ae94f44a2f5f7438306fe9e84fcd19

app/views/users/mailer/confirmation_instructions.html.erb
LinqLover:
```suggestion
         confirmation_url(@resource, :confirmation_token => @resource.confirmation_token) %></p>

```

app/views/users/mailer/reset_password_instructions.html.erb
LinqLover:
```suggestion
<p><%= t('.instruction_3', :default => "Your password won't change until you access the link above and create a new one.") %></p>

```

app/views/users/shared/_links.erb
LinqLover:
```suggestion
<% end -%>

```

config/locales/devise.en.yml
LinqLover:
```suggestion
        last_sign_in_at: "Last signed in at"

```

LinqLover:
> I find this PR really hard to review. Would it also be possible to customize only some views of devise?

Hard for the same reason because then we would mix up the old and the new style and could not rely on the autogeneration any longer. But I promise you that I really did not make any manual changes in cf9060f :-)

felixauringer (approved):

felixauringer:
> @felixauringer Thanks for your review! All your suggestions but the first one refer to auto-generated code (I did not write it by hand ...). Do you think it is a good idea to change these auto-generations? Otherwise, I'd just leave it as-is.

Fine for me, only some red tests to work on :+1:

PR #161, opened 2021-01-29
Added notifications

### LINKED ISSUE #128: [Process Part] Notifications for Contacts Invitation
# Description
Als Benutzer möchte ich Kontaktanfragen empfangen & beantworten können, um neue Kontakte zu erhalten.

# Dependencies
- [ ] Nutzer können Kontaktanfragen senden

# Conditions of satisfaction
- [ ] im Header gibt es ein Notification-Symbol, was mittels Badges anzeigt, wie viele eingegangene, unbearbeitete Kontaktanfragen ich habe
- [ ] wenn ich auf das Notification-Symbol klicke komme ich zu einer Seite (wie unten abgebildet), auf der ich die Komtaktanfragen beantworten kann.

# Zugehöriges Wireframe

Receive Invites

### LINKED ISSUE #205: Bell button in top-bar doesnt work
The bell button should navigate to my requests page

rheaSPK:
image

Closes #128


app/views/layouts/_navbar.html.erb
```diff
@@ -1,9 +1,16 @@
...
 <nav class="navbar navbar-expand-md navbar-dark bg-dark">
   <div class="navbar-brand"><%= link_to I18n.t('index.portal_name'), root_path, class: 'portal-link' %></div>
-  <form class="form-inline">
-    <i class="fa fa-bell" style="color: white; padding: 10px"></i>
-    <a href="<%= search_users_path %>"><i class="fa fa-search" style="color: white; padding: 10px"></i> </a>
-  </form>
+  <% if user_signed_in? %>
+    <form class="form-inline">
+      <% if current_user.contact_requests.any? %>
+        <a href="<%= user_contact_requests_path(current_user) %>"> <i class="fa fa-bell" style="color: red; padding: 3px"></i> </a>
```
felixauringer, line 7: I would extract all styles into an appropriate css file

app/views/layouts/_navbar.html.erb
```diff
@@ -1,9 +1,16 @@
 <%# https://getbootstrap.com/docs/4.5/components/navbar/ %>
 <nav class="navbar navbar-expand-md navbar-dark bg-dark">
   <div class="navbar-brand"><%= link_to I18n.t('index.portal_name'), root_path, class: 'portal-link' %></div>
-  <form class="form-inline">
-    <i class="fa fa-bell" style="color: white; padding: 10px"></i>
-    <a href="<%= search_users_path %>"><i class="fa fa-search" style="color: white; padding: 10px"></i> </a>
-  </form>
+  <% if user_signed_in? %>
+    <form class="form-inline">
+      <% if current_user.contact_requests.any? %>
```
felixauringer: Does that mean that the bell is not shown if the user does not have any notifications?

app/views/layouts/_navbar.html.erb
```diff
@@ -1,9 +1,16 @@
...
+        <a href="<%= user_contact_requests_path(current_user) %>"> <i class="fa fa-bell" style="color: red; padding: 3px"></i> </a>
+        <span class="badge badge-pill badge-danger"> <%=  current_user.contact_requests.length %> </span>
```
felixauringer:
```suggestion
        <span class="badge badge-pill badge-danger"><%= current_user.contact_requests.length %></span>
```
This code is really hard to understand because the relation between notifications and contact requests is not really clear. I would prefer to have the logic in the model rather than in the controller so that I can just call `current_user.notification_count` in the view. The `if` statement has the same problem, maybe `current_user.notifications?` would be nice there. This logic would then also need model specs.

app/views/layouts/_navbar.html.erb
```diff
@@ -1,9 +1,16 @@
...
+      <% else %>
+        <i class="fa fa-bell" style="color: white; padding: 10px"></i>
+      <% end %>
+      <a href="<%= search_users_path %>"><i class="fa fa-search" style="color: white; padding: 10px"></i> </a>
+    </form>
+  <% end %>
 <%= render 'layouts/navbar_desktop_part' %>
 </nav>
```
felixauringer: missing newline

felixauringer (changes requested):

app/views/layouts/_navbar.html.erb
rheaSPK: No, then there's a white bell without a link (it's in the `else` statement)

felixauringer: Oh, sorry :sweat_smile:

app/views/layouts/_navbar.html.erb
rheaSPK: I wouldn't like to do that, because if someone else uses the bell icon, they don't necessarily want my styling...

app/views/layouts/_navbar.html.erb
rheaSPK: Good point!

app/views/layouts/_navbar.html.erb
felixauringer: Yes but you could use a class like `notification-bell` or something like that. I just don't like to have HTML, CSS and application logic all in one file :sweat_smile: But if you prefer the inline styling that's fine for me because it's not such a big issue :+1:

frcroth: I agree with @felixauringer. If someone doesn't like the styling, the class can be changed even easier

app/models/user.rb
```diff
@@ -60,6 +60,14 @@ def init
     self.username ||= email.split('@', 2)[0]
   end
 
+  def has_contact_request?
```
felixauringer, line 63:
```suggestion
  def notifications?
```

app/models/user.rb
```diff
@@ -60,6 +60,14 @@ def init
...
+    contact_requests.any?
+  end
+
+  def count_contact_requests
```
felixauringer:
```suggestion
  def notification_count
```

felixauringer (commented):
My thought was to have descriptive methods in the model. We want to check whether the user has notifications, so I would name the method `notifications?`. This also makes it easier to change the defintion of a notification without touching the view again

PR #163, opened 2021-01-29
Feature: Make social handles hidable

LinqLover:
Part of #47. See https://github.com/hpi-swt2/connections-portal/issues/47#issuecomment-769821068.


LinqLover:
We decided on dropping this feature for the reason of limited capacities. It's a pity, but ...

Issue #160, opened 2021-01-29
Navigation problems


frcroth:
Mobile:

- [x] What does the 4th button (the fa-history button) do / what is planned here?
- [x] Proposals and activities should be removed
- [x] Double navbar is a little weird
- [x] Chat and home are selected at the same time
- [x] Chat in the lower navbar does not lead to chat
- Chat is not scrolled down (#178)

Desktop:

- [x] Contacts and requests are not visible
- [x] Proposals and activities should be removed
- [x] There is no way to get to my own page
- [x] There is no way to return to dashboard when on a seperate page
- [x] There is no way to go to contacts

LinqLover:
IMO this should get a high priority, it's a massive regression of our product

rheaSPK:
This isn't a bug, these exact points were explicitly mentioned in the acceptance criterias. I'm not happy with this solution too, so yes, this should be fixed. Maybe we should keep the old navbar for desktop and add the new one only for mobile... But that's up to the POs

LinqLover:
I believe the POs would expect us not to give up our common sense, it's just their responsibility to steer and overview the entire product in great detail ... But however, @florian-papsdorf would be nice to provide an appropriate wireframe to solve this :-)

j-hellenberg:
We currently don't care about the desktop version at all.

Concerning the mobile version, I think the points Felix brought up are what's left to do to reach proper navigation following the wireframes in #138 (+ chat tab).
Refer to #138 as well to see what is planned in the history tab.

Alexander-Dubrawski:
There is still one bug with the navigation bar. If you click on the search icon the link of the settings page is expanding over the whole page. That means that if I click somewhere on the page I get redirected to the edit user page. A quick fix for that is to wrap the link and Icon into a label and then give the label and link the following styling: 

```css
    .nav-bar-entry {
      position: relative;
    }
```

and the link should get the styling:

```css
    a::after {
        content: '';
        position: absolute;
        top: 0;
        left: 0;
        bottom: 0;
        right: 0;
    }
```

so for example the code could look as: 

```erb
...
      <li class="nav-item">
        <!-- please update the path when reworked the chat -->
          <label class="nav-bar-entry">
          <a class="nav-link <%= active_link_to_class(root_path, active: :exclusive)%>" href="<%= root_url %>">
            <i class="fa fa-comments"></i>
          </a>
        </label>
      </li>
...
```

This is just a small trick to style links correctly.

[...]

PR #167, opened 2021-02-01
Restore desktop navbar part

frcroth:
With the new navbar geared towards mobile, functionality for the desktop was removed. This PR reintroduces the drop down menu and sign out button removed in #147, it is only shown on the Desktop.
Don't worry product owners, this was done in my spare time and did not remove any time spent focusing on the mobile layout. This was just something I'd like personally.

### DIFF
--- a/app/assets/stylesheets/home.scss
+++ b/app/assets/stylesheets/home.scss
@@ -31,6 +31,10 @@ html, body {
   flex-wrap: wrap;
 }
 
+.portal-link {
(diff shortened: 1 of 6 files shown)


frcroth:
Would fix the following points from #160
- There is no way to get to my own page
- There is no way to return to dashboard when on a seperate page

sirkrypt0 (approved):
Nice

PR #168, opened 2021-02-02
Fix navbar on mobile

sirkrypt0:


frcroth:
Good that you noticed

PR #171, opened 2021-02-04
Two sided contacts

### LINKED ISSUE #84: Contacts are bidirectional
# Description
As a user, when i accepted the request of another user, i want him to be in my contact list

# Conditions of satisfaction
- [ ] when someone accepts a request, both user should have each other in their contacts

Paula-Kli:
closes #84 

everytime a contact in the database is created the inverse database entry is also created


SanJSp (approved):

PatriciaSowa (approved):

PatriciaSowa (approved):

PatriciaSowa (approved):

app/models/friendship.rb
```diff
@@ -0,0 +1,14 @@
+# A class for handling contacts, when one contact is created the other way is too
```
felixauringer: I think `when one contact is created the other way is too` isn't good English, not quite sure if my suggestion is better :sweat_smile: 
```suggestion
# A class for handling contacts which ensures that the association is always symmetrical
```

app/models/friendship.rb
```diff
@@ -0,0 +1,14 @@
...
+class Friendship < ApplicationRecord
+  self.table_name = 'users_users'
+  belongs_to :user, foreign_key: :contact_id, inverse_of: :friendships
+
+  after_create do |c|
```
felixauringer, line 9: These two callbacks should be tested, e.g. in a model spec

felixauringer: Something like this:

```ruby
require 'rails_helper'

RSpec.describe Friendship, type: :model do
  let(:user1) { FactoryBot.create :user }
  let(:user2) { FactoryBot.create :user }

  before { described_class.create(user_id: user1.id, contact_id: user2.id) }

  it 'creates a symmetrical association' do
    expect(described_class.find_by(contact_id: user1)).to be_present
    expect(described_class.find_by(contact_id: user2)).to be_present
  end

  it 'deletes both parts of the association' do
    described_class.find_by(contact_id: user1).destroy!
    expect(described_class.find_by(contact_id: user1)).to be_blank
    expect(described_class.find_by(contact_id: user2)).to be_blank
  end
end
```

felixauringer (changes requested):
Looks nice but I would like to have more tests for these fragile self-associations

app/models/user.rb
```diff
@@ -19,9 +19,8 @@ class User < ApplicationRecord
   # as we do not need to work with the relationship models as independent entities, `has_and_belongs_to_many` is fine
   # https://guides.rubyonrails.org/association_basics.html#choosing-between-has-many-through-and-has-and-belongs-to-many
   # rubocop:disable Rails/HasAndBelongsToMany
```
felixauringer, line 23: I would move this down so that it is directly above the code lines where this cop should be disabled

app/models/friendship.rb
felixauringer: I added these tests and some more

felixauringer (approved):

Issue #166, opened 2021-02-01
Show meeting history


j-hellenberg:
## User Story

**As a** user -  
**I want to** see which contacts I have recently called -  
**In order to** continue an interaction with them if they are available. 

## Dependencies
- #37

## Conditions of satisfaction

- [ ] the users I have recently called are shown, sorted by latest interaction
- [ ] the list is divided in users that are currently available (have state of available or Nice2MeetU set) and those that aren't
- [ ] I can send the users that are currently available a call invitation similar to those on the dashboard page

## Implementation Note

Don't care too much about card design here, this is subject of another issue.

## Wireframe

![image](https://user-images.githubusercontent.com/47941549/106452506-46bc2f80-6488-11eb-8f41-f12ee726efa3.png)

frcroth:
We will not be able to this in the remaining time. See #200.

PR #177, opened 2021-02-04
Move chat to new page, remove second navbar

### LINKED ISSUE #160: Navigation problems

frcroth:
Fixes #160 (apart from non-clickable history, that is described in #166 and scrolling down in the chat that is addressed in #178)


frcroth:
image

sirkrypt0 (approved):

app/views/home/dashboard.html+mobile.erb
```diff
@@ -3,25 +3,8 @@
...
-        <a class="nav-link active" data-toggle="pill" href="#users">
-          <i class="fa fa-group"></i>
-        </a></li>
-      <li class="nav-item">
-        <a class="nav-link" data-toggle="pill" href="#chat">
-          <i class="fa fa-comments"></i>
-        </a></li>
-    </ul>
   </div>
 </div>
```
felixauringer: missing newline

config/routes.rb
```diff
@@ -19,4 +19,6 @@
 
   resources :activities, only: :create
   resources :jitsi_calls, only: :create
+
+  get '/chat', to: 'home#chat', as: 'chat'
```
felixauringer:
```suggestion
  get 'chat', to: 'home#chat'
```
I think this should be sufficient but we have to test it first

PR #193, opened 2021-02-06
Use profile picture hotfix in show user and edit profile settings

felixauringer:
There were other occurences of `user.avatar`. I now hopefully found and hotfixed them all.

### DIFF
--- a/app/views/users/edit_profile.html.erb
+++ b/app/views/users/edit_profile.html.erb
@@ -6,12 +6,7 @@
   <% end %>
 
   <p>
-    <% if @user.avatar.attached? %>
-      <%= image_tag @user.avatar, height: 160, width: 160, style: "margin-bottom: 10px; display: block; margin-left: auto; margin-right: auto;", class: "rounded-circle" %>
-    <% else %>
-      <img class="rounded-circle" height="160" width="160" style="margin-bottom: 10px; display: block; margin-left: auto; margin-right: auto;"
-           src="https://user-images.githubusercontent.com/48409110/107123576-8aef6b80-689e-11eb-9386-16eecd0ceb9e.png">
-    <% end %>
+    <%= render 'avatar', user: @user, size: 160, style: "margin-bottom: 10px; display: block; margin-left: auto; margin-right: auto;" %>
   </p>
 
   <p>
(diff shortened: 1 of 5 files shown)


app/views/users/edit_profile.html.erb
```diff
@@ -6,7 +6,12 @@
   <% end %>
 
   <p>
-    <%= image_tag @user.avatar, height: 160, width: 160, :style => "margin-bottom: 10px; display: block; margin-left: auto; margin-right: auto;", class: "rounded-circle" %>
+    <% if @user.avatar.attached? %>
+      <%= image_tag @user.avatar, height: 160, width: 160, style: "margin-bottom: 10px; display: block; margin-left: auto; margin-right: auto;", class: "rounded-circle" %>
+    <% else %>
+      <img class="rounded-circle" height="160" width="160" style="margin-bottom: 10px; display: block; margin-left: auto; margin-right: auto;"
+           src="https://user-images.githubusercontent.com/48409110/107123576-8aef6b80-689e-11eb-9386-16eecd0ceb9e.png">
+    <% end %>
```
frcroth: At this point a template for rendering of user icons would be nice to remove code duplication. It should have the user and the size of the avatar as templates.

frcroth: `<%= render avatar %>`

felixauringer: You're right, the goal of this PR was just to get rid of 500er errors. The whole avatar logic has to be renewed anyway.

felixauringer: But I'll have the template ready in a minute

felixauringer: > At this point a template for rendering of user icons would be nice to remove code duplication. It should have the user and the size of the avatar as templates.

What's with all the styling?

frcroth: Well you can use a string that represents the styling for now. I think in the future it would be better to move the styling to some div above the template.

frcroth: Or better yet a class of course. Inline style is the devil's handiwork.

frcroth (approved):

app/views/users/edit_profile.html.erb
felixauringer: Done

Issue #179, opened 2021-02-04
Chat should show user avatars


frcroth:
As I understand it, you can access the user avatar via user.avatar. This would make it possible for the chat to show the user avatar instead of a placeholder. For dynamic messages this is harder, a path like users/id/avatar would be nice if it doesn't exist yet.

- [x] Chat that is rendered on reload shows avatars
- [x] Chat that is generated via web socket shows avatars

By the way, the user list on the dashboard should show avatars as well

chrisma:
Related to #185?

PR #156, opened 2021-01-28
Implement configurable user avatars

PatriciaSowa:
Closes #144

PR #195, opened 2021-02-06
Store avatars in database

### LINKED ISSUE #182: Profile Picture does not survive

felixauringer:
Fixes #182 

I removed ActiveStorage because we do not have access to something like S3. This PR saves our avatars as binary objects in the database. That may not be the prettiest solution but it should work on heroku.  
I used [this article](https://gist.github.com/macek/610596) as a guide.

Todos for my next working day:

- [x] update tests
- [x] set default avatar on creation of user
- [x] reload the shown image after update using javascript (reloading the whole page would empty the other form on the same page)
- [x] maybe try to fix the migration history properly (so that rollbacks would be possible)
- [x] remove unnecessary gem
- [x] properly display errors


app/controllers/application_controller.rb
```diff
@@ -4,14 +4,6 @@ class ApplicationController < ActionController::Base
...
 
   before_action :set_variant
-  before_action :configure_permitted_parameters,
-                if: :devise_controller?
-
-  def configure_permitted_parameters
-    devise_parameter_sanitizer.permit(:account_update) do |u|
-      u.permit(:avatar, :email, :password, :password_confirmation, :current_password)
-    end
-  end
```
frcroth, line 14: What's the reason for this change?

app/controllers/avatars_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def upload
+    @avatar = Avatar.find_by(user_id: current_user.id) || Avatar.new
+    file = params[:avatar][:file]
+    @saved = @avatar.update(
+      file: file.read,
+      filename: file.original_filename,
+      filesize: file.size,
+      mime_type: file.content_type,
+      user_id: current_user.id
```
frcroth: Should this be via current_user or rather via params user_id?

spec/models/avatar_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  it 'is not valid without a filename' do
+    expect(FactoryBot.build(:avatar, filename: nil)).not_to be_valid
+  end
+
+  it 'is not valid without a filesize' do
+    expect(FactoryBot.build(:avatar, filesize: nil)).not_to be_valid
+  end
+
+  it 'is not valid if it is bigger than 5MiB' do
```
frcroth, line 22: Is it megabytes or mibibyte?

app/views/avatars/upload.js.erb
```diff
@@ -0,0 +1,28 @@
...
+    <% end %>
+    const target = document.getElementById('new_avatar');
+    if (target.firstChild.id === 'error_explanation') {
+        target.replaceChild(error_node, target.firstChild);
+    } else {
+        target.insertBefore(error_node, target.firstChild);
+    }
+    generatePopup("<%= I18n.t('denial.resource_update', resource: Avatar) %>", 'fa-exclamation', [], 3000);
+  }
+<% end %>
```
frcroth, line 28: Don't let @sirkrypt0 see this, or he will turn it all into jQuery

app/views/avatars/upload.js.erb
```diff
@@ -0,0 +1,28 @@
...
+  {
+    const error_node = document.createElement('div');
+    error_node.id = 'error_explanation';
+    const list = document.createElement('ul');
+    error_node.appendChild(list);
+    let list_item;
+    <% @avatar.errors.full_messages.each do |message| %>
+      list_item = document.createElement('li');
+      list_item.textContent = '<%= message %>';
+      list.appendChild(list_item)
```
frcroth:
```suggestion
      list.appendChild(list_item);
```

spec/models/avatar_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    expect(FactoryBot.build(:avatar, filesize: 5.megabytes + 1)).not_to be_valid
+  end
+
+  it 'is not valid with a rational filesize' do
```
frcroth: What's a rational file size?

app/models/avatar.rb
```diff
@@ -0,0 +1,11 @@
+# A model for storing user avatars in the database
+class Avatar < ApplicationRecord
+  VALID_MIME_TYPES = %w[image/png image/jpg image/jpeg image/gif].freeze
+  MAX_FILE_SIZE_MEGABYTES = 5
+
+  validates :filename, :filesize, :file, :mime_type, presence: true
```
frcroth:
```suggestion
  validates :filename, :file_size, :file, :mime_type, presence: true
```
And everywhere else

app/controllers/application_controller.rb
felixauringer: This was added in the [PR](https://github.com/hpi-swt2/connections-portal/pull/156/files) that introduced profile pictures. I think that everything but the avatar is permitted by default because devise needs it. And because we do not need the avatar anymore I completely removed it to fall back to the default behaviour again.

app/controllers/avatars_controller.rb
felixauringer: There shouldn't be a `params[:user_id]` because avatars are not a subresource of the user. But because one should only change the profile picture of oneself, `current_user` is also the desired behaviour here.

spec/models/avatar_spec.rb
felixauringer: it does not return `5,000,000` but a little bit more, so I think it is actually M**e**biByte.

Seems like the [rails naming](https://api.rubyonrails.org/classes/Numeric.html#method-i-megabyte) isn't right here

app/views/avatars/upload.js.erb
felixauringer: I know this is ugly but I didn't know how to do it better. If jQuery is the way to go, then he can of course refactor this ^^

spec/models/avatar_spec.rb
felixauringer: A non-integer one, what would be the appropriate wording to describe this?

app/models/avatar.rb
felixauringer: Should I do the same for the `filename`?

frcroth: No, filename is an English word https://www.oxfordlearnersdictionaries.com/definition/english/filename?q=filename

frcroth:
Is there a way to access the avatars easily via a link that is predictable without ruby (for the js part of the global chat)? For #179

spec/models/avatar_spec.rb
frcroth: 5,243e+6 Bytes is 5 Mebibytes, if this is the returned value, change all occurrences of megabytes to mebibytes.

spec/models/avatar_spec.rb
frcroth: Ah no I get it, you mean as in rational number

frcroth:
```suggestion
  it 'is not valid with a non integer filesize' do
```

spec/models/avatar_spec.rb
felixauringer: `.megabytes` is a rails method, I could write an alias to that and name it `mebibytes` but I'm not sure if that would be a good solution. I could not find other (non-method calls) occurences of it in my code, did you?

felixauringer:
> Is there a way to access the avatars easily via a link that is predictable without ruby (for the js part of the global chat)? For #179

As the id of the avatar is unrelated to anything else, that is not possible. The easieast way I can think of is providing the link to the avatar together with the rest of the information (message, user, ...) via websocket. In my opinion, building links in javascript shouldn't be done anyway because of the many places one would have to change something if the link format would change.

app/models/avatar.rb
felixauringer: Done

spec/models/avatar_spec.rb
felixauringer: I did find one (the constant in the avatar model) and changed it there

spec/models/avatar_spec.rb
felixauringer: That's much clearer with your suggestion now

frcroth (approved):

konradh (approved):

konradh:
> > Is there a way to access the avatars easily via a link that is predictable without ruby (for the js part of the global chat)? For #179
> 
> As the id of the avatar is unrelated to anything else, that is not possible. The easieast way I can think of is providing the link to the avatar together with the rest of the information (message, user, ...) via websocket. In my opinion, building links in javascript shouldn't be done anyway because of the many places one would have to change something if the link format would change.

We moved the avatars to `/users/#{id}/avatar`.

app/views/users/edit_profile.html.erb
```diff
@@ -1,17 +1,23 @@
 <h2><%= I18n.t('user.edit.edit_profile_settings_heading') %></h2>
 
-<%= form_with(model: @user, local: true) do |form| %>
-  <% if @user.errors.any? %>
-    <%= render 'layouts/flash_messages'%>
-  <% end %>
+<p>
+  <%= render 'avatar', user: @user, size: 160, classes: %i[rounded-circle centered-avatar] %>
```
Paula-Kli, line 4: should `centered-avatar` be added to the `card.html` as well?

Paula-Kli (approved):

app/views/users/edit_profile.html.erb
felixauringer: It wasn't there before so I assumed that it is not necessary there. In my opinion it could be omitted completely as it does not look good on the profile page but this PR was more about making it work again, not making it pretty :sweat_smile:

PR #201, opened 2021-02-08
Chat avatars

### LINKED ISSUE #179: Chat should show user avatars

### LINKED ISSUE #185: Integrate Profile Pictures in global chat
The Icon on the side of sent messages should be replaced with the profile picture of the user that sent the message

konradh:


app/controllers/home_controller.rb
```diff
@@ -1,30 +1,28 @@
...
+  end
+
+  def filter_users_status
+    @users = User.where.not(id: current_user.id).with_status(User.filter_status)
+    @users = @users.sample(maximum_length_user_list)
+  end
+
   def maximum_length_user_list
     30
   end
```
frcroth: This will result in a merge conflict. But you should be able to handle it. maximum_length_user_list is a field now.

app/views/home/_chat.html.erb
```diff
@@ -19,11 +19,7 @@
 <div class="d-none" data-role="message-template">
   <div class="chat-item d-flex flex-row">
     <div class="avatar-container d-flex align-items-end">
-      <a href="#">
-        <span class="w-48 avatar">
-          <i class="fa fa-user"></i>
-        </span>
-      </a>
+      <%= render 'users/avatar', user: current_user, size: 48, classes: %i[rounded-circle] %>
```
frcroth: Is rounded-circle necessary here? Shouldn't all occurrences of avatar be circles?

spec/features/global_chat_spec.rb
```diff
@@ -35,17 +30,29 @@ def post_message(message)
...
+    end
+
+    it "displays the user's avatar" do
+      expect(page).to have_css("img[src='#{avatar_user_path(user)}']")
+    end
+
+    it 'shows the correct timestamp after reload' do
+      visit root_path
+      expect(page).to have_text(room_message.formatted_time)
+    end
```
frcroth, line 68: Add a test that it links to the user profile

spec/features/global_chat_spec.rb
```diff
@@ -35,17 +30,29 @@ def post_message(message)
...
+    it 'clears text box after message submit' do
+      expect(page).to have_field('room_message_message', text: '')
+    end
+
+    it 'shows the correct timestamp' do
+      expect(page).to have_text(room_message.formatted_time)
+    end
+
+    it "displays the user's avatar" do
+      expect(page).to have_css("img[src='#{avatar_user_path(user)}']")
```
frcroth: Use `within('#chat-messages')` to make it more specific

frcroth (changes requested):

app/views/home/_chat.html.erb
felixauringer: If that is the case, it can be moved to the avatar partial

spec/features/global_chat_spec.rb
```diff
@@ -35,17 +30,29 @@ def post_message(message)
     end
   end
 
-  it 'shows the correct timestamp' do
-    post_message(message)
-    message = RoomMessage.all.last
-    expect(page).to have_text(message.formatted_time)
-  end
+  context 'with a posted message' do
+    let(:room_message) { RoomMessage.all.last }
```
felixauringer: Why do we need this extra variable? Shouldn't the last message equal the outer `let(:message)`?
And if this is to check whether the message was saved, then I would add a new test for that rather than checking it implicitly by getting the last message from the database

spec/features/global_chat_spec.rb
felixauringer: Could be added to some more tests here

spec/features/global_chat_spec.rb
```diff
@@ -35,17 +30,29 @@ def post_message(message)
...
+
+    it 'shows the correct timestamp' do
+      expect(page).to have_text(room_message.formatted_time)
+    end
+
+    it "displays the user's avatar" do
+      expect(page).to have_css("img[src='#{avatar_user_path(user)}']")
+    end
+
+    it 'shows the correct timestamp after reload' do
```
felixauringer, line 63: I would check for all things (message, time, avatar, avatar link, name and name link) after posting and after reloading. Should be easy using shared examples and only a different before block. Can also be moved to another PR if you would prefer that

felixauringer (changes requested):

spec/features/global_chat_spec.rb
konradh: The outer message is a String, but in this tests we need to call `formatted_time` on a RoomMessage object.

felixauringer: Hmm, I see. Confusing names here, maybe one could name the outer one `message_text`, `message_string` or something like that?

spec/features/global_chat_spec.rb
```diff
@@ -35,17 +30,42 @@ def post_message(message)
...
+    let(:room_message) { RoomMessage.all.last }
 
-  it 'shows the correct timestamp after reload' do
-    post_message(message)
-    message = RoomMessage.all.last
-    visit root_path
-    expect(page).to have_text(message.formatted_time)
+    before do
+      post_message(message)
+      room_message
```
felixauringer: `let!` would do that for you. Then the before block could be a one liner with `{}`
EDIT: nope, it wouldn't because it is executed before :sweat_smile: Sorry

spec/features/global_chat_spec.rb
```diff
@@ -35,17 +30,42 @@ def post_message(message)
...
+      expect(page).to have_field('room_message_message', text: '')
+    end
+
+    it 'shows the correct timestamp' do
+      within('#chat-messages') do
+        expect(page).to have_text(room_message.formatted_time)
+      end
+    end
+
+    it 'links to the users profile page' do
```
felixauringer: We would need two more specific tests for that because the user name (do we test if that is only displayed for messages from other users?) and the profile picture both have the same link.

With so many changes to the tests, we could really open a new PR (if you did not start with the tests yet)

app/views/users/_avatar.html.erb
```diff
@@ -1,2 +1,4 @@
 <%= stylesheet_link_tag 'users', media: 'all', 'data-turbolinks-track': 'reload' %>
-<%= image_tag avatar_user_path(user), id: "avatar-#{user.id}", height: size, width: size, class: "#{classes.join(' ')}" %>
+<%= link_to user_path(user), class: "avatar-user-link" do %>
+  <%= image_tag avatar_user_path(user), id: "avatar-#{user.id}", height: size, width: size, class: "#{classes.append('avatar-image', 'rounded-circle', 'centered-avatar').join(' ')}" %>
```
felixauringer: With all these classes now in here, it would be possible to remove `classes` completely and just set it statically, wouldn't it?

felixauringer (approved):
I'm happy with the amount of tests now. There can always be more but I think it is fine now :+1:

app/controllers/home_controller.rb
felixauringer: This is resolved properly (by you, so I just resolve this :sweat_smile: )

frcroth (approved):
Merge it!

Issue #203, opened 2021-02-08
Remove user list


felixauringer:
The user list (`UsersController#index` / `/users`) looks ugly and isn't linked anywhere, so we should remove it. One can still see all users on the user search page which, by default, displays all users but the current_user.

PR #211, opened 2021-02-08
Fix button to add contact

### LINKED ISSUE #196: Clicking the '+' does not create contact request
When clicking on the '+' sign in the users search the other user does not get a contact request.

### LINKED ISSUE #204: Clicking the + navigates to my requests
When I'm searching for someone and I want to send him a contact-request (clicking the '+'), I don't want to be redirected to my personal requests page. I would expect to stay on the Search-Page

felixauringer:
Closes #204 
Closes #196

### DIFF
--- a/app/controllers/contact_requests_controller.rb
+++ b/app/controllers/contact_requests_controller.rb
@@ -3,10 +3,10 @@ class ContactRequestsController < ApplicationController
   before_action :find_user, only: [:accept]
 
   def create
-    requested_user = User.find(params[:user_id])
-    requested_user.contact_requests << current_user
-    requested_user.save
-    redirect_to users_path
+    @requested_user = User.find(params[:user_id])
+    return if @requested_user.contact_requests.include? current_user
+    @requested_user.contact_requests << current_user
+    @saved = @requested_user.save
   end
 
   def index
--- a/app/controllers/contact_requests_controller.rb
+++ b/app/controllers/contact_requests_controller.rb
@@ -5,6 +5,7 @@ class ContactRequestsController < ApplicationController
   def create
     @requested_user = User.find(params[:user_id])
     return if @requested_user.contact_requests.include? current_user
+    
     @requested_user.contact_requests << current_user
     @saved = @requested_user.save
   end
--- a/app/controllers/contact_requests_controller.rb
+++ b/app/controllers/contact_requests_controller.rb
@@ -5,7 +5,7 @@ class ContactRequestsController < ApplicationController
   def create
     @requested_user = User.find(params[:user_id])
     return if @requested_user.contact_requests.include? current_user
-    
+
     @requested_user.contact_requests << current_user
     @saved = @requested_user.save
   end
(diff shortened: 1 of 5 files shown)


frcroth:
Maybe you can fix #203 in this PR, as it is somewhat related

app/views/users/_card.html.erb
```diff
@@ -13,9 +13,12 @@
...
     <% if mode == :add %>
-      <label class="button-label">
-        <%= link_to (fa_icon 'plus-circle', title: I18n.t('user.contact_request.send')), user_contact_requests_path(user), id: user.id, class: "custom-link add_button btn btn-secondary"%>
-      </label>
+        <div class="col-1">
+            <%= form_with url: user_contact_requests_path(user), remote: true, id: "add-contact-form-#{user.id}" do |f| %>
+              <%= button_tag(type: 'submit', class: 'fa fa-3x fa-plus-circle user-contact-add-button', title: I18n.t('user.contact_request.send')) do %>
+              <% end %>
+            <% end %>
+          </div>
```
Alexander-Dubrawski:
```suggestion
      <%= form_with url: user_contact_requests_path(user), remote: true, id: "add-contact-form-#{user.id}" do |f| %>
        <%= button_tag(type: 'submit', class: 'fa fa-3x fa-plus-circle add_button', title: I18n.t('user.contact_request.send')) do %>
        <% end %>
      <% end %>
```
I would not use the `col-1` class. It seems to have a negative impact on the layout because it should be used to style tables in bootstrap:

Screenshot 2021-02-09 at 11 51 34

With the changes above and the adjustment of the `add_button` class inside the `users.scss`: 

```css
.add_button {
  background: none;
  border-width: 0;
  color: #dd640c;
}
```

The layout will look as: 

Screenshot 2021-02-09 at 11 59 41

Alexander-Dubrawski (changes requested):
Thanks for the great work. I just found one layout issue. I suggested some change that should fix it.

app/controllers/contact_requests_controller.rb
```diff
@@ -3,10 +3,11 @@ class ContactRequestsController < ApplicationController
   before_action :find_user, only: [:accept]
 
   def create
-    requested_user = User.find(params[:user_id])
-    requested_user.contact_requests << current_user
-    requested_user.save
-    redirect_to users_path
+    @requested_user = User.find(params[:user_id])
+    return if @requested_user.contact_requests.include? current_user
+
```
frcroth:
```suggestion

```

app/views/users/_card.html.erb
Paula-Kli: user-contact-add-button does exactly what you suggested as `add-button` but we can take out the diff

thanks for the hint with the styling... might happen due to the merge earlier today

Paula-Kli: I applied the hint with the div and I think the styling should be fixed now

felixauringer:
Approved :heavy_check_mark:

felixauringer:
> Maybe you can fix #203 in this PR, as it is somewhat related

@frcroth We decided not to do this here in order to get this fix merged quickly.

